PR #2, opened 2024-06-23
[PDS] Add CI Workflow

DrewMcArthur:
- adds `rust-toolchain`, which specifies a nightly rust version (2024-06-23)
- adds `.github/workflows/rust.yml`, which does a `cargo check, build, test` in that order
- fixes a couple of build errors

You can see the CI working on [the PR in my fork](https://github.com/DrewMcArthur/rsky/pull/1)

(oh also, let me know if you want me to rebase this onto `main` instead of this feature branch!)


.github/workflows/rust.yml
```diff
@@ -0,0 +1,51 @@
...
+  #     uses: actions/checkout@v4
+
+  #   - name: Rust Cache
+  #     uses: Swatinem/rust-cache@v2
+
+  #   - name: Run Clippy
+  #     run: cargo clippy  -- -D warnings
+
+  #   - name: Run cargo fmt
+  #     run: cargo fmt -- --check
```
DrewMcArthur, line 51: this fails, let me know if you would rather just delete it for now

rsky-firehose/src/main.rs
```diff
@@ -146,9 +146,7 @@ async fn process(message: Vec<u8>, client: &reqwest::Client) {
                                                 author: commit.repo.to_owned(),
                                                 record: post
                                             };
-                                            if let Some(prev) = commit.prev {
-                                                create.prev = Some(prev.to_string());
-                                            }
+                                            create.prev = commit.prev.clone();
```
DrewMcArthur, line 149: could probably move this up into the initialization above; I took out the `if` since the variable we're setting is an `Option` too

rsky-lexicon/src/com/atproto/sync.rs
rudyfraser, line 1: When I try this locally I get `the trait bound 'CidGeneric<64>: serde::Deserialize<'_>' is not satisfied [E0277]`. I prefer to use the Cid type for cid values whenever possible but found that I get this issue with serde. Do you see the same on your end?

rudyfraser (changes requested):
Most of this looks good to me! See my comment about `rsky-lexicon/src/com/atproto/sync.rs` which is the sole concern I have as I get errors for that change on my side.

rsky-lexicon/src/com/atproto/sync.rs
DrewMcArthur: i do not, I made this change because I was getting the following error from `cargo check`:

```
error[E0308]: mismatched types
   --> rsky-firehose/src/main.rs:136:84
    |
136 | ...                   let record_reader = Cursor::new(car_blocks.get(&cid).unwrap());
    |                                                                  --- ^^^^ expected `&Cid<64>`, found `&String`
    |                                                                  |
    |                                                                  arguments to this method are incorrect
    |
    = note: expected reference `&cid::cid::Cid<64>`
               found reference `&std::string::String`
```

note, my cargo/rustc version is: `cargo 1.81.0-nightly (bc89bffa5 2024-06-22)`.  I rolled back this file with `git checkout origin/pds -- rsky-lexicon/src/com/atproto/sync.rs`.

DrewMcArthur: ah! I was able to get your error by bumping up a few versions (`cid` to `0.11.1` and `serde_ipld_dagcbor` to `0.6.1`).  hm.

rudyfraser: Sounds good! I also know that rsky-firehose and rsky-feedgen use a different Cid library than rsky-pds and rsky-lexicon.
The PDS uses the cid from libipld which I think is the right one to use.
I'll open an issue for fixing the firehose code and using libipld if you don't mind holding off on making changes to rsky-firehose and rsky-feedgen for now.

DrewMcArthur:
closed in favor of #10

PR #11, opened 2024-07-22
`use libipld::cid::Cid`

### LINKED ISSUE #9: Standardize on libipld::Cid where possible
Libipld is the Rust library for IPLD (https://github.com/ipld/libipld) which includes support for IPLD objects and is used in conjunction with serde_ipld_dagcbor.

Newer crates in the rsky repo use libipld::Cid but the first two (rsky-firehose and rsky-feedgen) use the cid crate. There are some slight differences between the two and I think it would be best to standardize on one, preferably the libipld one.

I don't think rsky-feedgen even uses cid even though it's listed as a dependency (can probably just be removed). The firehose crate will need its car.rs module updated.

DrewMcArthur:
closes #9

this appears to compile and pass checks, without having to do much other than swapping out the libraries.

### DIFF
--- a/rsky-feedgen/Cargo.toml
+++ b/rsky-feedgen/Cargo.toml
@@ -22,7 +22,6 @@ serde_cbor = "0.11.2"
 diesel = { version = "2.1.0", features = ["postgres"] }
 dotenvy = "0.15"
 chrono = "0.4.26"
-cid = { version = "0.10.1", features = ["serde-codec"] }
 regex = "1.8.4"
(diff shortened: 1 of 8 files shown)


DrewMcArthur:
@rudyfraser let me know if there's more needed to close this out, i could definitely be missing something.  i just added some tests to `rsky_firehose::car` to make sure stuff still ran, but i don't have the most solid understanding of what's going on there. so maybe they're not necessary, or just bad tests.  happy to add more / take those out, whatever you think ✌️

PR #40, opened 2025-01-24
Implement Jetstream Support

TheRipperoni:
This creates an additional service that can be used as an alternative to the firehose subscriber. Advantage being significantly less bandwidth consumption as well as cpu time. 

It also allows for customizing which lexicons you want to specifically query.


rsky-jetstream/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "rsky-jetstream"
+version = "0.1.0"
+authors = ["Rudy Fraser <him@rudyfraser.com>", "Ripperoni <email@ripperoni.com>"]
+description = "Jetstream subscriber service"
+license = "Apache-2.0"
+edition = "2021"
+publish = true
```
rudyfraser, line 8: Since it's a service I just wanna confirm -- do you definitely wanna publish as a crate?

rsky-jetstream/Dockerfile
rudyfraser, line 1: Any reason this couldn't be simplified to:

```
FROM rust

WORKDIR /usr/src/rsky
COPY . .

RUN cargo build --package rsky-jetstream

CMD ["sh", "-c", "cargo run --package rsky-jetstream"]
```

rsky-jetstream/Cargo.toml
TheRipperoni: My thought was it does provide the some handy defined structs and processing for the jetstream currently. I could go either way with it

rudyfraser (changes requested):
Couple small changes I'd like you to take a look at for main.rs but then otherwise should be good to go. Thanks for getting started on this!

rudyfraser:
Oh, one other thing. Because this is a jetstream-subscriber and not meant as a [jetstream implementation](https://github.com/bluesky-social/jetstream) could you rename the package as such? Wouldn't want there to be a namespace clash down the line if someone wanted to implement jetstream in the future for rsky.

rsky-jetstream/src/main.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+}
+
+#[tracing::instrument]
+#[tokio::main]
+async fn main() {
+    let default_subscriber_path = env::var("JETSTREAM_SERVER_ENDPOINT")
+        .unwrap_or("wss://jetstream1.us-west.bsky.network".into());
+    let wanted_collections = env::var("FILTER_PARAM")
+        .unwrap_or("wantedCollections=app.bsky.feed.post&wantedCollections=app.bsky.feed.repost&wantedCollections=app.bsky.graph.follow&wantedCollections=app.bsky.feed.like".into());
```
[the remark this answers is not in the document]
TheRipperoni: Will update it accordingly, as that sounds good

rsky-jetstream/Dockerfile
TheRipperoni: Forgot to mention, but the other advantage is it leads to a smaller docker image as we only keep the built executable this way

TheRipperoni:
> Oh, one other thing. Because this is a jetstream-subscriber and not meant as a [jetstream implementation](https://github.com/bluesky-social/jetstream) could you rename the package as such? Wouldn't want there to be a namespace clash down the line if someone wanted to implement jetstream in the future for rsky.

Updated to rename the package to rsky-jetstream-subscriber, as this make it much more clear what it does

rudyfraser (approved):

PR #42, opened 2025-01-28
BugFix:Token Validation

TheRipperoni:
Current token validation behavior causes tokens to always be expired as the token is measured in seconds, but the comparison uses microseconds.  This fixes that issue.

In addition, the bsky client doesn't allow for certain characters in the token value, so have removed those from possible values to allow for support.


TheRipperoni:
Saw it didn't like my format, will update

TheRipperoni:
Actually,I'm not sure what the problem is with it

rudyfraser:
Don't worry about the formatting. I'll hit everything with a `cargo fmt` after merge.
I'll review and let you know if there's any feedback.

rudyfraser (approved):

PR #46, opened 2025-02-04
PDS:  Already Existing DID Create Account Support

TheRipperoni:
Redid the CreateAccount API call to be easier to read, as well  as have better tracing and error handling.

Removed the, now legacy, genesis creation methods as they are no longer needed.

Added necessary support for already existing DIDs. This step is critical in order to allow for account migration to RSky PDS

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ resolver = "2"
 cargo = { version = "0.84.0",features = ["vendored-openssl"] }
 rsky-lexicon = {path = "rsky-lexicon", version = "0.2.3"}
 rsky-identity = {path = "rsky-identity", version = "0.1.0"}
-rsky-crypto = {path = "rsky-crypto", version = "0.1.0"}
+rsky-crypto = {path = "rsky-crypto", version = "0.1.1"}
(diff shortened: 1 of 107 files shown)


rudyfraser (commented):
Can you elaborate on the changes related to the genesis operation? The genesis op itself uses a legacy format but the genesis block is still important and needs to be written to the did:plc log.

I'm not seeing where the genesis op is being sent to the plc directory in your new changes.

TheRipperoni:
> Can you elaborate on the changes related to the genesis operation? The genesis op itself uses a legacy format but the genesis block is still important and needs to be written to the did:plc log.
> 
> I'm not seeing where the genesis op is being sent to the plc directory in your new changes.

I moved the logic of the Genesis operation to the plc/operation instead. If you check line 344 of create_account.rs, you can see there is a call to the new create function inside plc which handles the creation of the Operation. Line 85 is where the create_account.rs invokes the plc_client to send the operation to the plc directory.

The logic of this was to separate the plc logic more to the plc package, and less to the create_account. It also doesn't invoke the plc call in the cases where a DID is supplied, as in that particular case the PLC already exists. That instead is used before invoking the importRepo call for migrating accounts

rudyfraser (approved):

PR #51, opened 2025-02-06
PDS: Fix catcher error handling

TheRipperoni:
Current behavior when request guards are triggered returns a http status of 500. This changes the behavior to respond as is intended by utilizing ApiError.

Technical note, the use of request.local_cache is to allow for us to pass the error response back to the catcher. Rocket has plans to add behavior to pass error bodies to the catcher in the future, but it is not currently supported otherwise.


TheRipperoni:
Merged latest main in

rudyfraser (approved):

PR #57, opened 2025-02-10
 PDS: Implemented Import Repo API

TheRipperoni:
Implemented functionality to import your repo to RSky PDS.

Reasons users may want to do that include but are not limited to:
1.  A user is migrating to a given RSky PDS
2. A user has already migrated, but has determined some outside data they would like to be included in their repo from another source


rsky-pds/src/apis/com/atproto/repo/import_repo.rs
```diff
@@ -1,4 +1,157 @@
...
+    let curr_repo: Option<Repo> = match curr_root {
+        None => None,
+        Some(_root) => Some(Repo::load(actor_store.storage.clone(), curr_root).await?),
+    };
+
+    // Process imported car
+    let import_datastream = blob.open(100.megabytes());
+    let import_bytes = import_datastream.into_bytes().await.unwrap().value;
+
+    let car_with_root = match read_car_with_root(import_bytes).await {
```
afbase, line 48: `CarReader::new` should be able to take a stream and this _might_ (I think I'd have to take a deeper look into the call stack) make the overhead of `into_bytes()` a little lighter for large blobs, yes?
 
 To take advantage of it, I think we'd need to tweak `rsky-repo/src/car.rs` so that:
 ```rust
 pub async fn read_car(bytes: Vec<u8>) -> Result<CarToBlocksOutput> {
    read_stream_car(bytes.as_slice()).await
}

pub async fn read_car_with_root(bytes: Vec<u8>) -> Result<CarWithRoot> {
    read_stream_car_with_root(bytes.as_slice()).await
}

pub async fn read_stream_car_with_root<R: AsyncRead + Send + Unpin>(bytes: R) -> Result<CarWithRoot> {
    let CarToBlocksOutput { roots, blocks } = read_stream_car(bytes).await?;
    if roots.len() != 1 {
        bail!("Expected one root, got {}", roots.len());
    }
    let root = roots[0];
    Ok(CarWithRoot { root, blocks })

}

pub async fn read_stream_car<R: AsyncRead + Send + Unpin>(bytes: R)  -> Result<CarToBlocksOutput> {
    let car = CarReader::new(bytes).await?;
    car_to_blocks(car).await
}
```

Then we'd have:
```rust
    // Process imported car
    let import_datastream = blob.open(100.megabytes());
    let car_with_root = match read_stream_car_with_root(import_datastream).await {
        Ok(res) => res,
        Err(error) => {
            return Err(ApiError::InvalidRequest(error.to_string()));
        }
    };
```

What do you think?

TheRipperoni: I like it, it also makes the handler logic a lot cleaner to understand. 

I'm going to change the 100 to an environment variable as well. Currently, atproto-PDS sets the import repo size limit by the blob size limit. This has issues because, for instance, my repo is 80 MB which is over it. I think it would make sense for the rsky-PDS to have a custom option for repo size imports

TheRipperoni: If that doesn't make it into this pull request specifically though, I think we can always have that be a follow up one

TheRipperoni: Since I'm adopting this, going to write the unit tests for it first

TheRipperoni: Updated accordingly and tested

rsky-pds/src/apis/com/atproto/repo/import_repo.rs
```diff
@@ -1,4 +1,157 @@
...
+
+    // Get current repo if it exists
+    let curr_root: Option<Cid> = actor_store.get_repo_root().await;
+    let curr_repo: Option<Repo> = match curr_root {
+        None => None,
+        Some(_root) => Some(Repo::load(actor_store.storage.clone(), curr_root).await?),
+    };
+
+    // Process imported car
+    let max_import_size = env_int("IMPORT_REPO_LIMIT").unwrap_or(100).megabytes();
```
afbase, line 46: Per [the documentation]:

>Import a repo in the form of a CAR file. Requires Content-Length HTTP header to be set.

It might take a little bit of work but one could fail early in the request if:
1. the content-length header is not set, 
2. the content-length header value is not a positve integer
3. the content-legnth header value exceeds `IMPORT_REPO_LIMIT` or 100 megabyes.  

and then it continues with the 
```rust
let requester = auth.access.credentials.unwrap().did.unwrap();
    let mut actor_store = ActorStore::new(
        requester.clone(),
        S3BlobStore::new(requester.clone(), s3_config),
        db,
    );
// code continues with rest of implementation ....
```

This would allow the server to ignore oversized requests right away without processing the entire stream of data.

TheRipperoni: I'll have to check, because from my own testing on my atproto-pds it didn't block me on content-length header. Do we want to utilize it to stay inline with the documentation, even if the atproto-pds doesn't follow suit?

TheRipperoni: Nice catch. I'll update it today accordingly as I agree

TheRipperoni: Updated and tested

rsky-pds/src/apis/com/atproto/repo/import_repo.rs
```diff
@@ -1,4 +1,201 @@
...
+                Ok(content_length) => {
+                    if content_length.bytes() > max_import_size {
+                        let error = ApiError::InvalidRequest(format!(
+                            "Content-Length is greater than maximum of {max_import_size}"
+                        ));
+                        req.local_cache(|| Some(error.clone()));
+                        return Outcome::Error((Status::BadRequest, error));
+                    }
+
+                    let import_datastream = data.open(content_length.megabytes());
```
afbase, line 52:
```suggestion
                    let import_datastream = data.open(content_length.bytes());
```

rsky-pds/src/apis/com/atproto/repo/import_repo.rs
```diff
@@ -1,4 +1,201 @@
...
+                Outcome::Error((Status::BadRequest, error))
+            }
+            Some(res) => match res.parse::<usize>() {
+                Ok(content_length) => {
+                    if content_length.bytes() > max_import_size {
+                        let error = ApiError::InvalidRequest(format!(
+                            "Content-Length is greater than maximum of {max_import_size}"
+                        ));
+                        req.local_cache(|| Some(error.clone()));
+                        return Outcome::Error((Status::BadRequest, error));
```
afbase, line 49:
```suggestion
                        return Outcome::Error((Status::PayloadTooLarge, error));
```

rsky-pds/src/apis/com/atproto/repo/import_repo.rs
```diff
@@ -1,4 +1,201 @@
...
+    type Error = ApiError;
+
+    #[tracing::instrument(skip_all)]
+    async fn from_data(req: &'r Request<'_>, data: Data<'r>) -> Outcome<'r, Self, Self::Error> {
+        let max_import_size = env_int("IMPORT_REPO_LIMIT").unwrap_or(100).megabytes();
+        match req.headers().get_one(header::CONTENT_LENGTH.as_ref()) {
+            None => {
+                let error = ApiError::InvalidRequest("Missing content-length header".to_string());
+                req.local_cache(|| Some(error.clone()));
+                Outcome::Error((Status::BadRequest, error))
```
afbase, line 40:
```suggestion
                Outcome::Error((Status::LengthRequired, error))
```

rsky-pds/src/apis/com/atproto/repo/import_repo.rs
```diff
@@ -1,4 +1,201 @@
...
+            }
+            Some(res) => match res.parse::<usize>() {
```
afbase, line 42: I know this is a little pedantic but perhaps worthwhile using the [NonZeroU64](https://doc.rust-lang.org/std/num/type.NonZeroU64.html).  

I have an example in the [PR](https://github.com/TheRipperoni/rsky/pull/1/files#diff-a2938bfecc087337c1c0394c07fc5308d2aa8e8ac237cd1af4d6aa5b3dc402d6R80) I made.

afbase (changes requested):

rsky-pds/src/apis/com/atproto/repo/import_repo.rs
TheRipperoni: Current typescript implementation from BSky ATProto sets the status to 400 (BadRequest or Invalid Request), while PayloadTooLarge would be 413. Do you want to deviate here? It also doesn't match the http documentation for bsky

rsky-pds/src/apis/com/atproto/repo/import_repo.rs
TheRipperoni: Current typescript implementation from BSky ATProto sets the status to 400 (BadRequest or Invalid Request), while LengthRequired would be 411. Do you want to deviate here? It also doesn't match the http documentation for bsky

rsky-pds/src/apis/com/atproto/repo/import_repo.rs
afbase: Ah wow 😰 

That is a bit of a pity to not have error specificity and we're stuck to 400, 401's with json message string to distinguish what went wrong.

Let's not deviate and stick to the API docs on errors.

rsky-pds/src/apis/com/atproto/repo/import_repo.rs
afbase: see my comment above - https://github.com/blacksky-algorithms/rsky/pull/57#discussion_r1955514872

rsky-pds/src/apis/com/atproto/repo/import_repo.rs
TheRipperoni: Can do, will make the update when I get home

TheRipperoni: Updated accordingly

afbase (approved):
This looks good to me!   



I don't know where we're at with spans and the tracing library but perhaps we can look at spans + publishing performance metrics over endpoints like importRepo using tracing + opentelemetry.

PR #58, opened 2025-02-10
 PDS: Implemented Submit PLC Operation API

TheRipperoni:
Allows users to use RSky PDS to submit PLC Operations that are given to it.

Reasons for this would include, but are not limited to:
1. A user is migrating to RSky PDS
2. A user wants to change something in their PLC DidDocument, but RSky PDS doesn't actually support the change they want to make natively.  This allows them to still use it to make the change they want


rsky-pds/src/apis/com/atproto/identity/submit_plc_operation.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+#[rocket::post(
+    "/xrpc/com.atproto.identity.submitPlcOperation",
+    format = "json",
+    data = "<body>"
+)]
+#[tracing::instrument(skip_all)]
+pub async fn submit_plc_operation(
+    body: Json<Operation>,
```
rudyfraser, line 112: Not the correct input type. Should be a new SubmitPlcOperationInput type in rsky-lexicon that has an `operation` field of type JsonValue

rsky-pds/src/apis/com/atproto/identity/submit_plc_operation.rs
```diff
@@ -0,0 +1,147 @@
...
+    auth: AccessStandard,
+    sequencer: &State<SharedSequencer>,
+    id_resolver: &State<SharedIdResolver>,
+    server_config: &State<ServerConfig>,
+) -> Result<(), ApiError> {
+    let did = auth.access.credentials.unwrap().did.unwrap();
+    let op = body.into_inner();
```
rudyfraser, line 119: Should instead be serde_json::from_value::<Operation>() and return invalid request if failing

rsky-pds/src/apis/com/atproto/identity/submit_plc_operation.rs
```diff
@@ -0,0 +1,147 @@
...
+    let public_endpoint = server_config.service.public_url.as_str();
+
+    validate_submit_plc_operation_request(did.as_str(), &op, public_endpoint).await?;
+
+    let plc_url = env_str("PDS_DID_PLC_URL").unwrap_or("https://plc.directory".to_owned());
```
rudyfraser, line 124: Should use cfg.identity.plc_url

rsky-pds/src/apis/com/atproto/identity/submit_plc_operation.rs
```diff
@@ -0,0 +1,147 @@
...
+    let plc_client = plc::Client::new(plc_url);
```
rudyfraser, line 125: Should use cfg.identity.plc_url

rudyfraser (changes requested):
See comments

rsky-pds/src/apis/com/atproto/identity/submit_plc_operation.rs
TheRipperoni: Will add

TheRipperoni: Presuming as prior to bump the version

TheRipperoni: Also, rather than solely JsonValue, if we follow the typescript implementation they actually strictly define some of the fields and types in the input

TheRipperoni: I am thinking to define it a bit stricter in the lexicon though, because the typescript implementation defines it fairly strictly at their schema level

`if (!check.is(op, plc.def.operation)) {
        throw new InvalidRequestError('Invalid operation')
      }

operation: z.ZodObject<z.extendShape<{
        type: z.ZodLiteral<"plc_operation">;
        rotationKeys: z.ZodArray<z.ZodString, "many">;
        verificationMethods: z.ZodRecord<z.ZodString, z.ZodString>;
        alsoKnownAs: z.ZodArray<z.ZodString, "many">;
        services: z.ZodRecord<z.ZodString, z.ZodObject<{
            type: z.ZodString;
            endpoint: z.ZodString;
        }, "strip", z.ZodTypeAny, {
            type: string;
            endpoint: string;
        }, {
            type: string;
            endpoint: string;
        }>>;
        prev: z.ZodNullable<z.ZodString>;
    }
`

rudyfraser: See my [other comment](https://github.com/blacksky-algorithms/rsky/pull/58#discussion_r1957127303). Should be JsonValue in lexicon and then "validated" using the turbofish operator `serde_json::from_value::<Operation>(value)`

rsky-pds/src/apis/com/atproto/identity/submit_plc_operation.rs
TheRipperoni: This differs from the other API endpoint implementations which utilize into_inner. Is there a particular reason for it here?

rudyfraser: into_inner() is fine. this comment is dependent on the [other change](https://github.com/blacksky-algorithms/rsky/pull/58#discussion_r1957126802) about the input type being changed. After you change the input type to SubmitPlcOperationInput, you'll need to validate the JsonValue from body.operation

To be very explicit, I'm suggesting something like:
```rust
#[tracing::instrument(skip_all)]
pub async fn submit_plc_operation(
    body: Json<SubmitPlcOperationInput>,
    auth: AccessStandard,
    sequencer: &State<SharedSequencer>,
    id_resolver: &State<SharedIdResolver>,
    server_config: &State<ServerConfig>,
) -> Result<(), ApiError> {
    let did = auth.access.credentials.unwrap().did.unwrap();
    let SubmitPlcOperationInput {
        operation
    } = body.into_inner();
    let operation = serde_json::from_value::<Operation>(operation).map_err(ApiError::BadRequest(String::from("InvalidRequest"),String::from("Invalid operation")))?;
// Rest of code
```

TheRipperoni: Ah, understood now. Much appreciated

rsky-pds/src/apis/com/atproto/identity/submit_plc_operation.rs
TheRipperoni: Updated

rudyfraser (approved):

PR #59, opened 2025-02-10
PDS: Implemented Sign PLC Operation API

TheRipperoni:
This implements the Sign PLC Operation API endpoint. This is necesary for any case where a user wants us to sign a given PLC operation, but not necessarily use it right away. Examples include, but are not limited to, if a user wants to migrate away from our PDS

### DIFF
--- a/rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
+++ b/rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
@@ -8,17 +8,8 @@ use crate::plc::operations::create_update_op;
 use crate::plc::types::{CompatibleOp, CompatibleOpOrTombstone, Operation, Service};
 use rocket::serde::json::Json;
 use rsky_common::env::env_str;
+use rsky_lexicon::com::atproto::identity::SignPlcOperationRequest;
 use std::collections::BTreeMap;
-
-#[derive(Clone, Debug, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
-pub struct SignPlcOperationRequest {
-    pub token: String,
-    pub rotation_keys: Vec<String>,
-    pub also_known_as: Vec<String>,
-    pub verification_methods: BTreeMap<String, String>,
-    pub services: BTreeMap<String, Service>,
-}
 
 #[rocket::post(
     "/xrpc/com.atproto.identity.signPlcOperation",
--- a/rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
+++ b/rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
@@ -56,15 +56,59 @@ pub async fn sign_plc_operation(
     let private_key = env_str("PDS_PLC_ROTATION_KEY_K256_PRIVATE_KEY_HEX").unwrap();
     let (secret_rotation_key, _) = get_keys_from_private_key_str(private_key)?;
 
+    //If request doesn't contain field, check last op for field. In the case of CreateOpV1,
+    // we don't set it (which is aligned with BSky Implementation
+    let also_known_as = match request.also_known_as {
+        None => match last_op {
+            CompatibleOp::CreateOpV1(_) => None,
+            CompatibleOp::Operation(ref op) => Some(op.also_known_as.clone()),
+        },
+        Some(res) => Some(res),
+    };
+    let services = match request.services {
+        None => match last_op {
+            CompatibleOp::CreateOpV1(_) => None,
+            CompatibleOp::Operation(ref op) => Some(op.services.clone()),
+        },
+        Some(res) => match serde_json::from_value::<BTreeMap<String, Service>>(res) {
+            Ok(services) if !services.is_empty() => Some(services),
+            _ => match last_op {
+                CompatibleOp::CreateOpV1(_) => None,
+                CompatibleOp::Operation(ref op) => Some(op.services.clone()),
+            },
+        },
+    };
+    let verification_methods = match request.verification_methods {
+        None => match last_op {
+            CompatibleOp::CreateOpV1(_) => None,
+            CompatibleOp::Operation(ref op) => Some(op.verification_methods.clone()),
+        },
+        Some(res) => Some(res),
+    };
+    let rotation_keys = match request.rotation_keys {
+        None => match last_op {
+            CompatibleOp::CreateOpV1(_) => None,
+            CompatibleOp::Operation(ref op) => Some(op.rotation_keys.clone()),
+        },
+        Some(res) => Some(res),
+    };
     let operation = match create_update_op(
         last_op,
         &secret_rotation_key,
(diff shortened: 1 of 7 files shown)


rudyfraser (changes requested):
The build for this is failing. See build logs for details.

TheRipperoni:
1. Ran recommended steps before Pull Request
2. Corrected Errors in build

TheRipperoni:
Errors caused by a missed commit

TheRipperoni (commented):
Updated build to no longer fail, as well as tested locally.

rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
```diff
@@ -0,0 +1,90 @@
...
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SignPlcOperationRequest {
+    pub token: String,
+    pub rotation_keys: Vec<String>,
+    pub also_known_as: Vec<String>,
+    pub verification_methods: BTreeMap<String, String>,
+    pub services: BTreeMap<String, Service>,
+}
```
rudyfraser, line 21: Move to rsky-lexicon/src/com/atproto/identity.rs
Bump rsky-lexicon version += 0.0.1

rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    let operation = match create_update_op(
+        last_op,
+        &secret_rotation_key,
+        |normalized: Operation| -> Operation {
+            let mut updated = normalized.clone();
+            updated.also_known_as = request.also_known_as.clone();
+            updated.services = request.services.clone();
+            updated.verification_methods = request.verification_methods.clone();
+            updated.rotation_keys = request.rotation_keys.clone();
```
rudyfraser, line 76: Typescript code at time of review is:
```typescript
        (lastOp) => ({
          ...lastOp,
          rotationKeys: input.body.rotationKeys ?? lastOp.rotationKeys,
          alsoKnownAs: input.body.alsoKnownAs ?? lastOp.alsoKnownAs,
          verificationMethods:
            // @TODO: actually validate instead of type casting
            (input.body.verificationMethods as
              | undefined
              | Record<string, string>) ?? lastOp.verificationMethods,
          services:
            // @TODO: actually validate instead of type casting
            (input.body.services as
              | undefined
              | Record<string, { type: string; endpoint: string }>) ??
            lastOp.services,
        }),
```
Does this need similar null-coalesce logic or is there a reason it doesn't? For example:
```rust
updated.also_known_as = if !request.also_known_as.is_empty() {
    request.also_known_as.clone()
} else {
    updated.also_known_as
};
```

rudyfraser (changes requested):
Mostly good. Couple small changes requested, see comments.

rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
TheRipperoni: Moving this to rsky0-lexicon causes a circular dependency as Service is defined in rsky-pds. If we want to move it, can we refactor to move that model into the rsky-lexicon then?

rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
TheRipperoni: No strong reason we don't have it. I can add this to it

rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
rudyfraser: Could use JsonValue instead of BTreeMap<String, T> which is my general preference when lexicon specifies 'unknown' for a type.

```rust
use std::collections::BTreeMap;
#[macro_use]
extern crate serde_derive;
use serde_json::{json, Value as JsonValue};

// From rsky-lexicon
#[derive(Clone, Debug, Serialize, Deserialize)]
#[serde(rename_all = "camelCase")]
pub struct SignPlcOperationRequest {
    pub token: String,
    pub rotation_keys: Vec<String>,
    pub also_known_as: Vec<String>,
    pub verification_methods: BTreeMap<String, String>,
    pub services: JsonValue,
}

// From rsky-pds
#[derive(Debug, Deserialize, Serialize, Clone)]
pub struct Service {
    #[serde(rename = "type")]
    pub r#type: String,
    pub endpoint: String,
}

#[derive(Debug, Deserialize, Serialize, Clone)]
pub struct Operation {
    #[serde(rename = "type")]
    pub r#type: String, // string literal `plc_operation`
    #[serde(rename = "rotationKeys")]
    pub rotation_keys: Vec<String>,
    #[serde(rename = "verificationMethods")]
    pub verification_methods: BTreeMap<String, String>,
    #[serde(rename = "alsoKnownAs")]
    pub also_known_as: Vec<String>,
    pub services: BTreeMap<String, Service>,
    // Omit<t.UnsignedOperation, 'prev'>
    pub prev: Option<String>,
    #[serde(skip_serializing_if = "Option::is_none")]
    pub sig: Option<String>,
}

fn main() {
    let request = SignPlcOperationRequest {
        token: String::from("test"),
        rotation_keys: vec![],
        also_known_as: vec![],
        verification_methods: BTreeMap::default(),
        services: json!({
            "atproto_pds": {
              "endpoint": "https://shiitake.us-east.host.bsky.network",
              "type": "AtprotoPersonalDataServer"
            }
        })
    };
    
    let mut operation = Operation {
        r#type: String::from("test"),
        rotation_keys: vec![],
        verification_methods: BTreeMap::default(),
        also_known_as: vec![],
        services: BTreeMap::default(),
        prev: None,
        sig: None
    };
    
    operation.services = match serde_json::from_value::<BTreeMap<String, Service>>(request.services) {
        Ok(services) if !services.is_empty() => services,
        _ => operation.services
    };
    println!("Operation {operation:?}");
}
```

TheRipperoni: Will update accordingly with this practice

TheRipperoni: Upon reviewing the code, the typescript implementation validates that the pattern of the request matches closely to service as visible in your other review comment. Should we still have it be that agnostic, as the value for the service isn't actually unkown

rudyfraser: For rsky-lexicon be agnostic using JsonValue. The match statement I've written here is doing the actual validation that is still a TODO for the Typescript.

TheRipperoni: Sounds good, in that case I can adopt that

rudyfraser (approved):

PR #68, opened 2025-03-11
Afbase/sync1.1 part 1

afbase:
This is going to be the first of many different PRs related to Sync1.1  

I think the PR strategy would be to do many small, reviewable ones instead of one big one.

As I make PRs in the future, I will address and document what I am fixing and how it conforms to the typescript implementation in #66.

## Files Changed

1. `rsky-lexicon/src/com/atproto/sync.rs`
2. `rsky-pds/src/actor_store/mod.rs`
3. `rsky-pds/src/actor_store/repo/mod.rs`
4. `rsky-pds/src/actor_store/repo/types.rs` (new file)
5. `rsky-pds/src/apis/com/atproto/admin/delete_account.rs`
6. `rsky-pds/src/apis/com/atproto/admin/update_account_handle.rs`
7. `rsky-pds/src/apis/com/atproto/server/delete_account.rs`
8. `rsky-pds/src/apis/com/atproto/sync/subscribe_repos.rs`
9. `rsky-pds/src/lexicon/lexicons.toml`
10. `rsky-pds/src/sequencer/events.rs`
11. `rsky-pds/src/sequencer/mod.rs`

## Summary

1. **Replaced Tombstone and Handle events with a new Sync event type**:
   - Removed `TombstoneEvt` and `HandleEvt` types
   - Added a new `SyncEvt` and `TypedSyncEvt` structure
   - `SubscribeReposTombstone` replaced with `SubscribeReposSync`

2. **Updated synchronization data structure**:
   - Added a new `SyncEvtData` structure containing repository metadata (CID, rev, blocks)
   - Added functionality to get sync event data via `get_sync_event_data()` method

3. **Modified account handling**:
   - Updated account deletion logic to remove tombstone event sequencing
   - Updated handle update logic, removing explicit handle update sequencing
   - Added a TODO comment for an incomplete branch in account handle updates

4. **Updated the event sequencing system**:
   - Modified the event deserialization logic to handle the new event types
   - Updated the request sequence range function to use the new event types

5. **Modified repository subscription mechanism**:
   - Updated the `subscribe_repos` function to handle the new sync events
   - Updated lexicon schemas in TOML configuration

The changes appear to be streamlining the repository synchronization process by consolidating multiple event types into a more comprehensive sync event that includes repository block data directly, rather than using separate events like tombstones and handle updates.

## Related Issues

Relates to #66 and the PRs:

- https://github.com/bluesky-social/atproto/commit/6e382f67aa73532efadfea80ff96a27b526cb178
[...]

PR #89, opened 2025-04-10
Afbase/pseudo list repos by collection

afbase:
## Summary

@utkarshgupta137

This might be useful to reference to implement [listReposBycollection](https://docs.bsky.app/docs/api/com-atproto-sync-list-repos-by-collection) for the relay which is where it is supposed to be implemented instead of the PDS.

My implementation took advantage of the [record_did_collection_idx](https://github.com/blacksky-algorithms/rsky/blob/main/rsky-pds/migrations/2023-11-15-004814_pds_init/up.sql#L133) index on the PDS.


afbase:
related to https://github.com/blacksky-algorithms/rsky/pull/68

PR #98, opened 2025-05-07
rsky-relay: fix false negatives for commit verification

utkarshgupta137:
## Summary
Some cases, which are allowed to pass-through in the indigo relay are currently getting blocked by our relay. Update the verification behavior to bring it inline with indigo relay.

## Changes
- [ ] Feature implementation
- [x] Bug fix
- [ ] Documentation update
- [ ] Other (please specify):

## Checklist
- [ ] I have tested the changes (including writing unit tests).
- [ ] I have formatted my code correctly

### DIFF
--- a/rsky-relay/src/validator/utils.rs
+++ b/rsky-relay/src/validator/utils.rs
@@ -46,7 +46,6 @@ pub fn verify_commit_event(commit: &SubscribeReposCommit, root: Cid, prev: &Repo
     if let Some(since) = &commit.since {
         if since != &prev.rev {
             tracing::debug!(%since, "commit with miss-matching since");
-            return false;
         }
     } else {
         // NOTE: some PDSs don't send this field, so we continue verifying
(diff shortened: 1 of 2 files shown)


rsky-relay/src/validator/utils.rs
```diff
@@ -91,12 +89,12 @@ pub fn verify_commit_event(commit: &SubscribeReposCommit, root: Cid, prev: &Repo
             Ok(inv) => {
                 if !inv {
                     tracing::debug!(%idx, ?op, "unable to invert op");
-                    return false;
+                    return true;
```
rudyfraser, line 92: Will approve this but just making a note to understand that the intention of this is to line up with the fact that Indigo is currently [lenient](https://docs.bsky.app/blog/relay-sync-updates) about doing the op inversion currently, right?

rudyfraser (approved):

rsky-relay/src/validator/utils.rs
utkarshgupta137: Yeah: https://github.com/bluesky-social/indigo/blob/main/atproto/repo/sync.go#L101-L107.
Even though they're returning an error, the [caller](https://github.com/bluesky-social/indigo/blob/main/cmd/relay/relay/verify.go#L153) just logs the error.

PR #105, opened 2025-05-14
Fix image deploy

afbase:
## Summary
1. update rsky-pds dockerfile to optimize layering trick the @TheRipperoni suggested
2. update rsky-pds dockerfile to use a debian running image inspired by the [rust dockerhub](https://hub.docker.com/_/rust) documentation
3. fix the cid package dependency mismatch in firehose:
```shell
❯  cargo check -p rsky-firehose
    Checking rsky-firehose v0.2.0 (/Users/clint/src/bsky-code/rsky/rsky-firehose)
error[E0308]: mismatched types
   --> rsky-firehose/src/main.rs:146:88
    |
146 | ...                   let record_reader = Cursor::new(car_blocks.get(&cid).unwrap());
    |                                                                  --- ^^^^ expected `&CidGeneric<64>`, found `&Cid<64>`
    |                                                                  |
    |                                                                  arguments to this method are incorrect
    |
note: two different versions of crate `cid` are being used; two types coming from two different versions of the same crate are different types even if they look the same
   --> /Users/clint/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cid-0.11.1/src/cid.rs:70:1
    |
70  | pub struct Cid<const S: usize> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this is the found type `cid::cid::Cid`
    |
   ::: /Users/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cid-0.10.1/src/cid.rs:67:1
    |
67  | pub struct Cid<const S: usize> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this is the expected type `CidGeneric`
    |
   ::: rsky-firehose/src/main.rs:5:5
    |
5   | use libipld::Cid;
    |     ------- one version of crate `cid` used here, as a dependency of crate `libipld_core`
6   | use rsky_lexicon::app::bsky::feed::like::Like;
    |     ------------ one version of crate `cid` used here, as a dependency of crate `rsky_lexicon`
    = help: you can use `cargo tree` to explore your dependency tree
note: method defined here
   --> /Users/user/.rustup/toolchains/nightly-2025-01-03-aarch64-apple-darwin/lib/rustlib/src/rust/library/std/src/collections/hash/map.rs:897:12
    |
897 |     pub fn get<Q: ?Sized>(&self, k: &Q) -> Option<&V>
    |            ^^^

For more information about this error, try `rustc --explain E0308`.
error: could not compile `rsky-firehose` (bin "rsky-firehose") due to 1 previous error
```

## Changes
- [x] Feature implementation
- [x] Bug fix

### DIFF
--- a/.github/workflows/ghcr.yml
+++ b/.github/workflows/ghcr.yml
@@ -46,7 +46,7 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Extract metadata (tags, labels) for Docker
+      - name: Extract metadata (tags, labels) for ${{ matrix.service.name }} Docker
         id: meta
         uses: docker/metadata-action@v4
         with:
@@ -56,7 +56,7 @@ jobs:
             type=ref,event=branch
             type=raw,value=latest,enable={{is_default_branch}}
 
-      - name: Build and push Docker image
+      - name: Build and push ${{ matrix.service.name }} Docker image 
         uses: docker/build-push-action@v4
         with:
           context: .
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ cargo = { version = "0.84.0",features = ["vendored-openssl"] }
 serde = { version = "1.0.160", features = ["derive"] }
 serde_derive = "^1.0"
 serde_ipld_dagcbor = {  version = "0.6.1" ,features = ["codec"]}
-lexicon_cid = { package = "cid", version = "0.11.1", features = ["serde-codec"] }
+lexicon_cid = { package = "cid", version = "0.10.1", features = ["serde-codec"] }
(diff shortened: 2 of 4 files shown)


rudyfraser (commented):
Any reason for those specific versions for Multihash and cid? Just wanna make sure I understand why we can't be on the latest version of those crates

afbase:
> Any reason for those specific versions for Multihash and cid? Just wanna make sure I understand why we can't be on the latest version of those crates

there is a dependency issue mismatch as `libipld` refers to older versions of cid and multihash.  an ipld maintainer mentioned [ipld-core](https://github.com/ipld/libipld/issues/197#issuecomment-2005582143) is the library to migrate to.

PR #145, opened 2025-12-19
Add `com.atproto.sync.getHostStatus` xrpc endpoint

uniphil:
## Summary

Since the entries from `com.atproto.sync.listHosts` are almost the same (a superset actually) of the host status info, implementing this is mostly a copy of that endpoint which just looks up a specific host.

Note that `listHosts` left the actual *status* field as a TODO since rsky-relay doesn't actually track host state -- this PR does **not** introduce a new host state tracking system. It hard-codes that response field as `"active"` just like `listHosts` did.

It's not completely useless though: an OK response indicates that at least `rsky-relay` does know about an upstream host.

I tested the changes locally and confirmed
- 400 error with message for missing `hostname` parameter
- 400 error with message for an unknown host
- 200 OK with json response for a known host

(no unit test added since there aren't tests set up in the repo)

## Changes
- [x] Feature implementation
- [ ] Bug fix
- [ ] Documentation update
- [ ] Other (please specify):

## Checklist
- [ ] I have tested the changes (including writing unit tests).
- [ ] I have formatted my code correctly

### DIFF
--- a/rsky-relay/src/server/server.rs
+++ b/rsky-relay/src/server/server.rs
@@ -25,7 +25,7 @@ use crate::config::{HOSTS_MIN_ACCOUNTS, HOSTS_RELAY};
 use crate::crawler::{RequestCrawl, RequestCrawlSender};
 use crate::publisher::{MaybeTlsStream, SubscribeRepos, SubscribeReposSender};
 #[cfg(not(feature = "labeler"))]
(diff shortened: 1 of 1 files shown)


afbase (approved):
LGTM!  thank you @uniphil!

PR #153, opened 2026-02-23
Iss152/release profile debug

### LINKED ISSUE #152: DevOps:  [profile.release] debug = 2 causes excessive memory usage and OOM failures in containerized builds
### Summary

The workspace `Cargo.toml` sets `debug = 2` in `[profile.release]`, which enables full debug symbols in all release builds. This causes significant memory pressure during linking — enough to OOM-kill the build process in containerized environments (Docker/Podman) — and produces unnecessarily large binaries for production deployments.

### Observed Behavior

Running `docker build` for `rsky-pds` fails with `signal: 9, SIGKILL` during the final link step of `rsky-pds` itself (after all dependencies have compiled successfully). The process is killed by the OS out-of-memory killer, not by a compiler error. This occurs even after dedicating 8GB of RAM and 6 vCPU to the build machine.

```
error: could not compile `rsky-pds` (lib)

Caused by:
  process didn't exit successfully: `rustc --crate-name rsky_pds ...` (signal: 9, SIGKILL: kill)
```

### Root Cause

`Cargo.toml` line 44:

```toml
[profile.release]
debug = 2
```

`debug = 2` (full debug info) combined with `opt-level = 3` (full optimization) means the compiler must process both fully-optimized IR and complete DWARF debug symbols simultaneously during codegen and linking. For a crate like `rsky-pds` with ~50 transitive dependencies, this creates a very high peak memory requirement at link time.

### Expected Behavior

Release builds intended for production deployment should not include debug symbols by default. Debug info is valuable for profiling and debugging, but should be opt-in.

### Proposed Fix

Remove `debug = 2` from `[profile.release]` and add a dedicated `profiling` profile for developers who need debug symbols in an optimized build. This follows the existing named-profile pattern already present in the workspace:

```toml
[profile.release]
# debug = 2 removed — use --profile profiling for symbolized builds

[profile.profiling]
inherits = "release"
debug = 2
```

Developers needing symbols for a profiler (e.g., samply, perf, Instruments) can build with:

```sh
cargo build --profile profiling -p rsky-pds
```

All other release builds — including CI and Docker — use `cargo build --release` and get lean, production-appropriate binaries.

### Benefits

- Fixes OOM failures in Docker/Podman builds with reasonable memory limits
- Reduces binary size for production deployments
- Preserves full profiling capability via explicit opt-in
- Consistent with the existing named-profile pattern (`wasm-dev`, `server-dev`, `android-dev`) already in `Cargo.toml`

### Affected File

`Cargo.toml` — workspace root, affects all crates built with `--release`

jmcauley:
## Summary

Fixes OOM (out-of-memory) failures when building `rsky-pds` in containerized environments
(Docker/Podman). Full debug symbols in `[profile.release]` caused the linker to be killed
by the OS during the `rsky-pds` link step due to peak memory pressure from combining
`opt-level=3` and `debuginfo=2` across ~50 transitive dependencies.

Also fixes a pre-existing bug in `rsky-pds/Dockerfile` where `COPY` source paths were
incorrect for the build context, causing the second build stage to always fail.

## Related Issues

Closes #152

## Changes

- [ ] Feature implementation
- [x] Bug fix
- [ ] Documentation update
- [ ] Other (please specify):

## Checklist

- [x] I have tested the changes (including writing unit tests).
  - Verified `docker build .` completes successfully from `rsky-pds/` with 8 GB container
    memory after both fixes. Previously failed with `SIGKILL` (OOM) at the `rsky-pds` link
    step.
  - N/A — build configuration change only, no protocol behavior affected.
  - N/A
- [x] I have formatted my code correctly
  - Developers needing symbolized builds for profiling (samply, perf, Instruments) can use:
    `cargo build --profile profiling -p rsky-pds`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,10 @@ rsky-repo = {path = "rsky-repo", version = "0.0.2"}
 rsky-firehose = {path = "rsky-firehose", version = "0.2.1"}
 
 [profile.release]
-debug = 2  # Or any level from 0 to 2
+
+[profile.profiling]
+inherits = "release"
+debug = 2
 
 [profile.wasm-dev]
 inherits = "dev"
--- a/rsky-pds/Dockerfile
+++ b/rsky-pds/Dockerfile
@@ -18,9 +18,9 @@ RUN mkdir -p rsky-pds/src && echo "fn main() {}" > rsky-pds/src/main.rs
 RUN cargo build --release --package rsky-pds
 
 # Now copy the real source code and build the final binary
-COPY rsky-pds/src rsky-pds/src
-COPY rsky-pds/migrations rsky-pds/migrations
-COPY rsky-pds/diesel.toml rsky-pds/diesel.toml
+COPY src rsky-pds/src
+COPY migrations rsky-pds/migrations
+COPY diesel.toml rsky-pds/diesel.toml
 
 RUN cargo build --release --package rsky-pds
 


afbase:
@jmcauley 

Would it be possible to revise this PR so that either _all_ dockerfiles have conformity with copying dataand the CI will build the dockerfiles + a second PR for the optimization parameters?

PR #167, opened 2026-03-29
fix: service auth token handling and did:web resolution

rabble:
## Summary

- **Service auth fallback to repo signing key**: When JWT key verification fails, try the repo signing key (`PDS_REPO_SIGNING_KEY_K256_PRIVATE_KEY_HEX`). This enables accepting service auth tokens from external services like `video.bsky.app` that sign tokens with the PDS's repo signing key rather than the JWT session key.
- **Handle missing scope/sub in service tokens**: Service auth tokens from video.bsky.app use `lxm` instead of `scope` and `iss` instead of `sub`. The verifier now defaults to `Access` scope when scope is empty and falls back to `iss` when `sub` is absent.
- **Graceful error handling**: Replace `unwrap()` calls on sub/aud with proper error messages instead of panics. Accept `PDS_ADMIN_PASSWORD` as alias for `PDS_ADMIN_PASS`.
- **did:web resolution**: Serve `/.well-known/did.json` for `did:web:{hostname}` resolution, deriving the secp256k1 public key multibase from the PDS repo signing key. This enables service-to-service auth where the PDS identity is `did:web:hostname`.

## Test plan

- [ ] Verify session tokens (JWT key) still work for normal auth flows
- [ ] Verify service auth callback from video.bsky.app is accepted (token signed with repo key, missing scope, iss instead of sub)
- [ ] Verify `GET /.well-known/did.json` returns valid DID document with correct public key
- [ ] Verify admin auth works with both `PDS_ADMIN_PASSWORD` and `PDS_ADMIN_PASS`
- [ ] Run `cargo check -p rsky-pds` to verify compilation

### DIFF
--- a/rsky-pds/Cargo.toml
+++ b/rsky-pds/Cargo.toml
@@ -30,6 +30,7 @@ email_address = "0.2.4"
 event-emitter-rs = "0.1.4"
 futures = "0.3.28"
 hex = "0.4.3"
+multibase = "0.9.1"
 image = "0.25.1"
 indexmap = { version = "1.9.3", features = ["serde-1"] }
 infer = "0.15.0"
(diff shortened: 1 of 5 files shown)


afbase:
relates to https://github.com/blacksky-algorithms/rsky/pull/171

PR #171, opened 2026-04-08
shifts JWT public key derivation from per-token-validation to on-startup

seanwatters:
## Summary

This PR takes the JWT public key derivation from the `PDS_JWT_KEY_K256_PRIVATE_KEY_HEX` env var that was spread across a couple of functions and being run per-request/validation, and shifts it up into a `static` initialization that is only run on startup. It also drops the `async` from two functions (`verify_jwt` + `validate_bearer_token`) which do not perform any async operations or contain any `.await`s.

The overall goal of this change is to reduce the amount of work on the server for each authenticated request.

## Changes
- [ ] Feature implementation
- [ ] Bug fix
- [ ] Documentation update
- [x] Other (please specify): perf optimization

## Checklist
- [x] I have tested the changes (including writing unit tests).
- [x] I confirm that my implementation aligns with the [canonical Typescript implementation](https://github.com/bluesky-social/atproto) and/or [atproto spec](https://atproto.com/specs/atp) TS ATproto is storing this caching the [`_jwt_key`](https://github.com/bluesky-social/atproto/blob/700074694157718e33c974b666f4563f9647a50c/packages/pds/src/auth-verifier.ts#L469) on the class instance, but because these functions in Rust are bare and not associated to a struct, using `LazyLock` as a `static` seemed like a reasonable compromize
- [ ] I have updated relevant documentation. - n/a
- [x] I have formatted my code correctly
- [ ] I have provided examples for how this code works or will be used - n/a

Super excited about the project and if this kind of submission isn't the kind of thing you're looking for I'd love to find another better way to help out!

### DIFF
--- a/rsky-pds/src/account_manager/helpers/auth.rs
+++ b/rsky-pds/src/account_manager/helpers/auth.rs
@@ -6,14 +6,13 @@ use diesel::*;
 use jwt_simple::prelude::*;
 use rsky_common::time::{from_micros_to_utc, MINUTE};
 use rsky_common::{get_random_str, json_to_b64url, RFC3339_VARIANT};
-use secp256k1::{Keypair, Message, SecretKey};
+use secp256k1::{Message, SecretKey};
 use sha2::{Digest, Sha256};
 use std::time::SystemTime;
 use thiserror::Error;
 
 pub struct CreateTokensOpts {
     pub did: String,
-    pub jwt_key: Keypair,
     pub service_did: String,
     pub scope: Option<AuthScope>,
     pub jti: Option<String>,
@@ -77,7 +76,6 @@ pub enum AuthHelperError {
 pub fn create_tokens(opts: CreateTokensOpts) -> Result<(String, String)> {
     let CreateTokensOpts {
         did,
-        jwt_key,
         service_did,
         scope,
         jti,
@@ -85,7 +83,6 @@ pub fn create_tokens(opts: CreateTokensOpts) -> Result<(String, String)> {
     } = opts;
     let access_jwt = create_access_token(CreateTokensOpts {
         did: did.clone(),
-        jwt_key,
         service_did: service_did.clone(),
         scope,
         expires_in,
@@ -93,7 +90,6 @@ pub fn create_tokens(opts: CreateTokensOpts) -> Result<(String, String)> {
     })?;
     let refresh_jwt = create_refresh_token(CreateTokensOpts {
         did,
-        jwt_key,
         service_did,
         jti,
         expires_in,
@@ -105,7 +101,6 @@ pub fn create_tokens(opts: CreateTokensOpts) -> Result<(String, String)> {
 pub fn create_access_token(opts: CreateTokensOpts) -> Result<String> {
     let CreateTokensOpts {
         did,
-        jwt_key,
         service_did,
         scope,
         expires_in,
@@ -121,16 +116,13 @@ pub fn create_access_token(opts: CreateTokensOpts) -> Result<String> {
     )
     .with_audience(service_did)
     .with_subject(did);
-    // alg ES256K
-    let key = ES256kKeyPair::from_bytes(jwt_key.secret_bytes().as_slice())?;
-    let token = key.sign(claims)?;
-    Ok(token)
+
+    crate::auth_verifier::JWT_KEY.sign(claims)
 }
 
 pub fn create_refresh_token(opts: CreateTokensOpts) -> Result<String> {
     let CreateTokensOpts {
         did,
-        jwt_key,
         service_did,
         jti,
         expires_in,
@@ -147,10 +139,8 @@ pub fn create_refresh_token(opts: CreateTokensOpts) -> Result<String> {
     .with_audience(service_did)
     .with_subject(did)
     .with_jwt_id(jti);
-    // alg ES256K
-    let key = ES256kKeyPair::from_bytes(jwt_key.secret_bytes().as_slice())?;
-    let token = key.sign(claims)?;
-    Ok(token)
+
+    crate::auth_verifier::JWT_KEY.sign(claims)
 }
 
 pub async fn create_service_jwt(params: ServiceJwtParams) -> Result<String> {
@@ -197,10 +187,10 @@ pub async fn create_service_jwt(params: ServiceJwtParams) -> Result<String> {
 }
 
 // @NOTE unsafe for verification, should only be used w/ direct output from createRefreshToken() or createTokens()
-pub fn decode_refresh_token(jwt: String, jwt_key: Keypair) -> Result<RefreshToken> {
-    let key = ES256kKeyPair::from_bytes(jwt_key.secret_bytes().as_slice())?;
-    let public_key = key.public_key();
-    let claims = public_key.verify_token::<CustomClaimObj>(&jwt, None)?;
+pub fn decode_refresh_token(jwt: String) -> Result<RefreshToken> {
+    let claims = crate::auth_verifier::JWT_KEY
+        .public_key()
+        .verify_token::<CustomClaimObj>(&jwt, None)?;
     assert_eq!(
         claims.custom.scope,
         AuthScope::Refresh.as_str().to_owned(),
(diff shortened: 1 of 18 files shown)


seanwatters:
@afbase, all great points!

made the requested changes and ran the tests locally and everything seems to pass.

i've noticed at least [one other](https://github.com/blacksky-algorithms/rsky/blob/560fbec000eef4be4623810860fbb1769508d89e/rsky-pds/src/actor_store/mod.rs#L343) area where keys are being pulled from env vars and then instantiated in the way the `JWT_KEY` was previously; would it be useful for me to go and find as many of those instances as i can and pull them out to their own one-time `static`/`LazyLock`s? (probably letting this PR land first)

afbase:
T

> @afbase, all great points!
> 
> made the requested changes and ran the tests locally and everything seems to pass.
> 
> i've noticed at least [one other](https://github.com/blacksky-algorithms/rsky/blob/560fbec000eef4be4623810860fbb1769508d89e/rsky-pds/src/actor_store/mod.rs#L343) area where keys are being pulled from env vars and then instantiated in the way the `JWT_KEY` was previously; would it be useful for me to go and find as many of those instances as i can and pull them out to their own one-time `static`/`LazyLock`s? (probably letting this PR land first)

That sounds good.  Will review tomorrow!

rsky-pds/src/auth_verifier.rs
```diff
@@ -22,6 +22,14 @@ use thiserror::Error;
 
 const INFINITY: u64 = u64::MAX;
 
+pub static JWT_KEY: std::sync::LazyLock<ES256kKeyPair> = std::sync::LazyLock::new(|| {
```
afbase: Could we rename this PDS_JWT_KEYPAIR to reduce ambiguity?

rsky-pds/src/account_manager/helpers/auth.rs
```diff
@@ -121,16 +116,13 @@ pub fn create_access_token(opts: CreateTokensOpts) -> Result<String> {
     )
     .with_audience(service_did)
     .with_subject(did);
-    // alg ES256K
-    let key = ES256kKeyPair::from_bytes(jwt_key.secret_bytes().as_slice())?;
-    let token = key.sign(claims)?;
-    Ok(token)
+
+    crate::auth_verifier::JWT_KEY.sign(claims)
```
afbase: can we add an import statement at the top of the file?

rsky-pds/src/account_manager/helpers/auth.rs
```diff
@@ -147,10 +139,8 @@ pub fn create_refresh_token(opts: CreateTokensOpts) -> Result<String> {
     .with_audience(service_did)
     .with_subject(did)
     .with_jwt_id(jti);
-    // alg ES256K
-    let key = ES256kKeyPair::from_bytes(jwt_key.secret_bytes().as_slice())?;
-    let token = key.sign(claims)?;
-    Ok(token)
+
+    crate::auth_verifier::JWT_KEY.sign(claims)
```
afbase: can we add an import statement at the top of the file?

afbase (changes requested):
I think we're close.  please incorporate the other change you mentioned and see my comments.

I really appreciate this!

afbase:
@seanwatters do you mind add the change for https://github.com/blacksky-algorithms/rsky/blob/560fbec000eef4be4623810860fbb1769508d89e/rsky-pds/src/actor_store/mod.rs#L343 ?

seanwatters:
> @seanwatters do you mind add the change for
> 
> https://github.com/blacksky-algorithms/rsky/blob/560fbec000eef4be4623810860fbb1769508d89e/rsky-pds/src/actor_store/mod.rs#L343
> 
> ?

@afbase, added changes — in previous note about that var i mentioned breaking them out into separate PRs but totally fine to do it on this one too.

one other thing that i found while i was moving `PDS_REPO` to a single spot: it looks like in `rsky-pds/src/apis/com/atproto/identity/submit_plc_operation.rs` the [`get_public_rotation_key`](https://github.com/blacksky-algorithms/rsky/blob/827a35be3ee527963c6e0d23f0a5958c9773f783/rsky-pds/src/apis/com/atproto/identity/submit_plc_operation.rs#L33) is never called, and — in what i think is [supposed to be its place](https://github.com/blacksky-algorithms/rsky/blob/827a35be3ee527963c6e0d23f0a5958c9773f783/rsky-pds/src/apis/com/atproto/identity/submit_plc_operation.rs#L95) — the `get_public_signing_key` fn is called, instead.

would also be happy to extract the `PDS_PLC_ROTATION_KEY_K256_PRIVATE_KEY_HEX` into its own static var the way we've done with the first two, but wanted to get clarification on that `get_public_signing_key`/`get_public_rotation_key` thing first.

afbase:
> but wanted to get clarification on that `get_public_signing_key`/`get_public_rotation_key` thing first.

Great find! yes we need to use the rotation key! Please adjust that and then yes extract `PDS_PLC_ROTATION_KEY_K256_PRIVATE_KEY_HEX` into its own static var.

Thank you!

seanwatters:
> yes we need to use the rotation key! Please adjust that and then yes extract `PDS_PLC_ROTATION_KEY_K256_PRIVATE_KEY_HEX` into its own static var.

@afbase, knocked those two out and removed some more of the helpers that became dead code post-extraction

rsky-pds/src/main.rs
```diff
@@ -2,6 +2,10 @@ use rsky_pds::build_rocket;
 
 #[rocket::main]
 async fn main() {
+    let _ = &rsky_pds::context::PDS_REPO_SIGNING_KEYPAIR;
+    let _ = &rsky_pds::auth_verifier::PDS_JWT_KEYPAIR;
+    let _ = &rsky_pds::apis::com::atproto::server::PDS_PLC_ROTATION_KEYPAIR;
```
afbase:
```suggestion
    let _ = &*rsky_pds::context::PDS_REPO_SIGNING_KEYPAIR;
    let _ = &*rsky_pds::auth_verifier::PDS_JWT_KEYPAIR;
    let _ = &*rsky_pds::apis::com::atproto::server::PDS_PLC_ROTATION_KEYPAIR;
```

the `&*` does  a proper deref since AFAIK LazyCell does not implement the deref trait.  or we could call [force](https://doc.rust-lang.org/beta/std/cell/struct.LazyCell.html#method.force)

afbase (changes requested):

rsky-pds/src/main.rs
afbase: I pinky promise this is the _last_ thing and then we merge

seanwatters: ah, no worries at all! i should've tested that to verify it was actually triggering the init there. good catch.

afbase:
Hey @seanwatters — great PR, the optimization direction is exactly right. I tracked down the build failure and wanted to explain what's happening before pushing a fix.

**Root cause: the Dockerfile clones `main` directly**

`rsky-pds/Dockerfile` has this on line 7:
```dockerfile
RUN git clone --depth 1 https://github.com/blacksky-algorithms/rsky.git .
```

It then only `COPY`s `rsky-pds/src` from the build context. This means `rsky-repo` is always sourced from the `blacksky-algorithms/rsky` main branch, not from the PR branch. So the `&Keypair` signature changes you made in `rsky-repo/src/repo.rs` and `rsky-repo/src/util.rs` are invisible to the Docker build — but the matching changes in `rsky-pds/src/actor_store/mod.rs` are picked up, causing a type mismatch at three call sites:

- `actor_store/mod.rs:111` — passes `keypair: &Keypair` to `Repo::format_init_commit`, which on main still takes `Keypair` by value
- `actor_store/mod.rs:146` — same
- `actor_store/mod.rs:343` — passes `&PDS_REPO_SIGNING_KEYPAIR` (`&LazyLock<Keypair>`) to `Repo::format_commit`, which expects `Keypair` by value

**The fix**

Fortunately `secp256k1::Keypair` is `Copy`, so we can just deref at those three call sites (`*keypair` and `*PDS_REPO_SIGNING_KEYPAIR`). This keeps your static-initialization approach intact while matching the by-value signatures in `rsky-repo` until those changes land in main.

Pushing the fix to your branch now.

afbase (approved):
LGTM

PR #175, opened 2026-04-09
Fix: increment postsCount by local writes in read-after-write profile munge

### LINKED ISSUE #164: PDS: Read-after-write profile munging preserves stale upstream `postsCount` after local writes
**Describe the bug**

After `com.atproto.repo.createRecord`, the local read-after-write path is inconsistent:
- `app.bsky.feed.getAuthorFeed` immediately includes the new post
- `app.bsky.actor.getProfile` still returns stale `postsCount`
- `app.bsky.actor.getProfiles` can time out in that window, and later still return stale `postsCount`

This causes a real client-visible failure: on `bsky.app`, `post -> own profile` can land on the correct profile URL but render only the shell/footer with no profile content. The same
sequence works against `bluesky-pds` on the same VPS.

**To Reproduce**

1. Log into `https://bsky.app` against `https://rsky.pdslab.net`
2. Create a post
3. Immediately navigate to `/profile/<own-handle>`
4. Observe that the page lands on the correct profile URL but the profile content never appears
5. Compare with `bluesky-pds`, where the same flow works

**Expected behavior**

After a local write, profile reads should reflect the updated local state consistently enough for `bsky.app` to render the profile page correctly.

**Details**

- Operating system: `Linux 6.12.63+deb13-arm64 aarch64 GNU/Linux`
- Node version: `v20.19.2`

**Additional context**

- Tested version: `main` at `d2cb2b4e695a6f25b5981cab21eb6c7621fe4de8`
- Rust: `rustc 1.94.0 (4a4ef493e 2026-03-02)`
- Direct local XRPC trace:
  - `createRecord`: `200` fast
  - `getAuthorFeed`: `200`, includes the new post immediately
  - `getProfile`: `200`, but stale `postsCount`
  - `getProfiles`: can time out immediately after the write; later returns `200` but still stale `postsCount`
- Files:
  - [get_profile.rs](rsky-pds/src/apis/app/bsky/actor/get_profile.rs)
  - [get_profiles.rs](rsky-pds/src/apis/app/bsky/actor/get_profiles.rs)
  - [viewer.rs](rsky-pds/src/read_after_write/viewer.rs)
- Exact code detail:
  - `get_profile_munge()` and `get_profiles_munge()` call `local_viewer.update_profile_detailed(...)`
  - [update_profile_detailed](rsky-pds/src/read_after_write/viewer.rs#L564) preserves `followers_count`, `follows_count`, and `posts_count` from the original upstream view unchanged,
while only overlaying profile-record fields like `display_name`, `description`, `avatar`, and `banner`
- Minimal fix:
  Make read-after-write profile munging patch derived profile counters from local records instead of preserving stale upstream counts unchanged. The likely change is in
[update_profile_detailed](rsky-pds/src/read_after_write/viewer.rs#L564) and the `get_profile_munge` / `get_profiles_munge` call sites.

peroumal1:
Fixes #164

## What

`update_profile_detailed` in `read_after_write/viewer.rs` now accepts a `local_posts_count: usize` parameter and adds it to the upstream `posts_count` via a new `apply_local_posts_count` helper. Both call sites (`get_profile_munge` and `get_profiles_munge`) pass `local.posts.len()`.

## Why

After `createRecord`, the read-after-write munge was overlaying profile-record fields (`displayName`, `description`, `avatar`, `banner`) but passing upstream `followers_count`, `follows_count`, and `posts_count` through unchanged. The stale `postsCount` caused `bsky.app` to render a blank profile shell immediately after a post, because the appview response was inconsistent with the local feed state.

## Testing

3 unit tests for `apply_local_posts_count` in `read_after_write::viewer::tests`:
- local posts are added to the upstream count
- zero local posts leave the count unchanged
- `None` upstream count stays `None` (unknown, not guessed)

Full `LocalViewer`-level integration testing is not feasible without a live database, which is noted as a known limitation.

## Checklist

- [x] Changes have been tested, including unit tests
- [x] Implementation aligns with the canonical TypeScript implementation and/or the atproto spec
- [x] Relevant documentation — n/a
- [x] Code is formatted (`cargo fmt`)
- [x] Examples: unit tests serve as usage examples

### DIFF
--- a/rsky-pds/src/apis/app/bsky/actor/get_profile.rs
+++ b/rsky-pds/src/apis/app/bsky/actor/get_profile.rs
@@ -96,7 +96,7 @@ pub fn get_profile_munge(
             if original.did != requester {
                 return Ok(original);
             }
-            Ok(local_viewer.update_profile_detailed(original, profile.record))
+            Ok(local_viewer.update_profile_detailed(original, profile.record, local.posts.len()))
         }
     }
 }
--- a/rsky-pds/src/apis/app/bsky/actor/get_profiles.rs
+++ b/rsky-pds/src/apis/app/bsky/actor/get_profiles.rs
@@ -92,7 +92,11 @@ pub fn get_profiles_munge(
                     if prof.did != requester {
                         prof
                     } else {
-                        local_viewer.update_profile_detailed(prof, profile.record.clone())
+                        local_viewer.update_profile_detailed(
+                            prof,
+                            profile.record.clone(),
+                            local.posts.len(),
+                        )
                     }
                 })
                 .collect::<Vec<ProfileViewDetailed>>();
--- a/rsky-pds/src/read_after_write/viewer.rs
+++ b/rsky-pds/src/read_after_write/viewer.rs
@@ -565,6 +565,7 @@ impl LocalViewer {
         &self,
         view: ProfileViewDetailed,
         record: Profile,
+        local_posts_count: usize,
     ) -> ProfileViewDetailed {
         let ProfileViewDetailed {
             did,
@@ -620,7 +621,7 @@ impl LocalViewer {
             },
             followers_count,
             follows_count,
-            posts_count,
+            posts_count: apply_local_posts_count(posts_count, local_posts_count),
             associated,
             joined_via_starter_pack,
             viewer,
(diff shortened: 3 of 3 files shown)


afbase:
isn't this going to double count the posts?

afbase:
I think the count between the appview and what the PDS here is going to get confused

peroumal1:
Hey,
You might be right, I'm looking for now for a way to test this further, will update the PR accordingly if needed.

peroumal1:
Ok so,  back with more tests! I was able to confirm the implementation won't double count the posts.

To validate it I built a mock AppView test harness for rsky-pds. The test spins up a full Rocket + Postgres instance and replaces the AppView HTTP endpoint with a lightweight mock server that returns a controlled response (like our case here). 

The integration test then:                                                                                                                                                                                                     1. Creates an account and writes a post locally via the PDS                                                                                                                                                       
2. Calls app.bsky.actor.getProfile while the AppView mock returns the stale count
3. Asserts that the munged response reflects postsCount + local_writes — confirming apply_local_posts_count fires correctly in the read-after-write path                                                                                                                                                                                                                                                                         

The testing logic lives [in a separate branch in my fork.](https://github.com/peroumal1/rsky/tree/feat/local-dev-docker-stack). Happy to discuss the testing approach further in a separate discussion, and open a PR for this as well if this is something that would be welcome.

afbase:
I think the approach to consider is to follow the atproto reference pds read-after-write implementation that doesn't account for postCounts.  An appview should eventually stabilize the post count.

PR #176, opened 2026-04-09
Fix: replace panicking unwrap() in datetime parsers

### LINKED ISSUE #20: PDS: Panic when making calls to com.atproto.server.refreshSession
Panicking during refresh session calls. Issue seems to stem from datetime parsing:
```
POST /xrpc/com.atproto.server.refreshSession? application/json:
   >> Matched: (refresh_session) POST /xrpc/com.atproto.server.refreshSession
 thread 'rocket-worker-thread' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/chrono-0.4.26/src/naive/datetime/mod.rs:122:18:
 invalid or out-of-range datetime
 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
   >> Handler refresh_session panicked.
    >> This is an application bug.
    >> A panic in Rust must be treated as an exceptional event.
    >> Panicking is not a suitable error handling mechanism.
    >> Unwinding, the result of a panic, is an expensive operation.
    >> Panics will degrade application performance.
    >> Instead of panicking, return `Option` and/or `Result`.
    >> Values of either type can be returned directly from handlers.
    >> A panic is treated as an internal server error.
    >> Outcome: Error(500 Internal Server Error)
    >> Responding with registered (default_catcher) default catcher.
```

peroumal1:
Fixes #20

## What

`from_str_to_micros` and `from_str_to_utc` in `rsky-common/src/time.rs` previously called `.unwrap()` on `NaiveDateTime::parse_from_str`, converting any format mismatch into a panic that crashed the Rocket worker thread.

Both functions now:
- Return `Result` instead of panicking
- Try the primary `%Y-%m-%dT%H:%M:%S%.3fZ` format first
- Fall back to `DateTime::parse_from_rfc3339` to handle `+00:00` offsets and other stored variants

**Caller updates:**
- `account_manager/mod.rs` — propagates with `?` (token rotation returns `Result`)
- `email_token.rs` — propagates with `?`
- `read_after_write/util.rs` — propagates with `?`
- `subscribe_repos.rs` — logs a warning and falls back to `Utc::now()` to avoid aborting the event stream

## Why

`refreshSession` panicked when `token.expires_at` from the database didn't exactly match the strict `%.3fZ` format. Any row written by a different tool or an older code path would crash the worker thread with `invalid or out-of-range datetime`.

## Testing

6 unit tests added in `rsky-common::time::tests`:
- Primary format parses without error
- RFC 3339 with `+00:00` offset resolves to the same instant as `Z` suffix
- Invalid inputs return `Err` (not a panic)
- All 18 existing `rsky-common` tests continue to pass

## Checklist

- [x] Changes have been tested, including unit tests
- [x] Implementation aligns with the canonical TypeScript implementation and/or the atproto spec
- [x] Relevant documentation — n/a
- [x] Code is formatted (`cargo fmt`)
- [x] Examples: unit tests serve as usage examples

### DIFF
--- a/rsky-common/src/time.rs
+++ b/rsky-common/src/time.rs
@@ -1,5 +1,5 @@
 use crate::RFC3339_VARIANT;
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use chrono::offset::Utc as UtcOffset;
 use chrono::{DateTime, NaiveDateTime, Utc};
 use std::time::SystemTime;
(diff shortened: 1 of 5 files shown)


afbase (approved):
LGTM.

There is a refactor to be had where we can use the diesel time typing and not go through string conversion all together but requires a bit of work.

PR #184, opened 2026-04-26
Eliminate AppView slow-tail 502s: rsky-graph + handle-resolution split

rudyfraser:
## Summary

- Refactor wintermute handle-resolution sweep to two index-friendly queries (priority-A NULL handles every iteration, priority-B stale non-NULL every 20th iteration) — eliminates the `OR` clause that defeated `actor_null_handle_idx`. Drops 980 ms × 20K calls/hour to <10 ms each.
- Replace rsky-graph's long-lived `DECLARE CURSOR` bulk loader (the previous failure mode that hurt prod) with two safer paths: file-based load via `\copy` CSV (zero PG transaction during load) and keyset-paginated short-transaction queries with throttling. New `--load-from-file` / `GRAPH_LOAD_FROM_FILE` flag.
- Wire follow deletes through the firehose by recording `(actor_uid, rkey) → subject_uid` on creates so deletes can resolve the subject.

## Context

HAR captures showed users hitting `502 UpstreamFailure` from the PDS pipethrough at ~10s — the AppView itself was processing 27% of requests in >5s due to a mutual-follow self-join (645 ms × 1.24 M calls/hour) and the wintermute handle-resolution sweep (980 ms × 20K calls/hour). Plan in `~/.claude/plans/recursive-honking-sprout.md`.

A separate atproto branch `fix/graph-service-fallback` adds the feature-flagged `BSKY_GRAPH_SERVICE_URL` path in `getFollowsFollowing`, default off, falls back to PG on any error.

The starter-pack count fix (partial index `idx_profile_joined_via_sp`) was already deployed to production PG separately — `CREATE INDEX CONCURRENTLY`, no code change needed. Query 565 ms → 6 ms confirmed.

## Test plan

- [x] `cargo fmt`, `cargo clippy --all-targets`, `cargo build --release` pass for rsky-graph and rsky-wintermute
- [ ] `sudo /opt/appview/deploy-wintermute.sh main` once merged — auto-rollback in script handles regressions
- [ ] rsky-graph rollout is a separate operational sequence (off-peak `\copy` of `bsky.follow`, systemd unit install, firehose-only validation, feature-flag ramp) — code lands here but stays dormant until that sequence runs

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1125,6 +1125,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0669d5a35b64fdb5ab7fb19cae13148b6b5cbdf4b8247faf54ece47f699c8cef"
 
 [[package]]
+name = "bit-vec"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d2c54ff287cfc0a34f38a6b832ea1bd8e448a330b3e40a50859e6488bee07f22"
+
+[[package]]
 name = "bit_field"
 version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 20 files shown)


rudyfraser:
Closing per review: rsky-graph code needs cleanup (empty config.rs, firehose pattern needs to mirror rsky-firehose/wintermute, wrong default relay host) and the branch carries unrelated rsky-relay changes from feat/rsky-graph. Will reopen after rebasing onto main with the cleaned code.

PR #187, opened 2026-05-11
rsky-graph: incremental periodic save + drop eager bloom rebuild

rudyfraser:
## Summary

rsky-graph maintains an in-memory follow graph (RoaringBitmap per user, persisted to LMDB) so mutual-followers queries are sub-millisecond instead of a Postgres self-JOIN. This change makes the periodic LMDB save O(dirty users) via a `DashSet<u32>` populated in `add_follow`/`remove_follow` and drained per save, and drops the eager bloom-filter rebuild from `load_from_lmdb` since blooms can be repopulated incrementally as `add_follow` runs.

## Test plan

- [ ] `cargo fmt -p rsky-graph` clean
- [ ] `cargo clippy -p rsky-graph --all-targets` no new warnings
- [ ] `cargo test -p rsky-graph --lib` passes (includes 6 new tests in `graph::tests` covering dirty-tracking semantics)
- [ ] `cargo test -p rsky-graph --test smoke` passes (includes new `incremental_save_only_writes_dirty_entries` round-trip)
- [ ] `cargo build --release -p rsky-graph` clean

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1125,6 +1125,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0669d5a35b64fdb5ab7fb19cae13148b6b5cbdf4b8247faf54ece47f699c8cef"
 
 [[package]]
+name = "bit-vec"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


afbase:
@rudyfraser look at #189 to help with the rsky-pds issue

PR #201, opened 2026-07-09
Feature/bunny cdn token auth

rishibalakrishnan:
Drop the referrer header as the security mechanism for video CDN access and instead add token auth. This fixes the issue on mobile where video playback doesn't pass that header.

URLs are signed with Bunny's URL Token Authentication (advanced/HS256, path-style directory tokens), so relatively-resolved variant playlists and `.ts` segments under `/{guid}/` inherit the token. The token TTL is derived from the redirect `Cache-Control` max-ages (longest max-age + 24h playback margin), so a cached 307 can never carry an expired token.

## Environment variables

Defaults are the **final** deployment values; the overrides exist for the phased cutover below.

| Variable | Default | Purpose |
|---|---|---|
| `BUNNY_LIBRARY_ID` | required | Bunny Stream library |
| `BUNNY_API_KEY` | required | Bunny Stream API key |
| `BUNNY_PULL_ZONE` | required | Pull zone subdomain (without `.b-cdn.net`) |
| `BUNNY_TOKEN_KEY` | required when `CDN_TOKEN_AUTH=true` | URL Token Authentication key (Stream → Library → Security) |
| `CDN_TOKEN_AUTH` | `true` | `false` = emit unsigned playback URLs (cutover only; logs a startup warning) |
| `PLAYLIST_REDIRECT_MAX_AGE_SECS` | `3600` | `Cache-Control` max-age of the playlist 307 |
| `THUMBNAIL_REDIRECT_MAX_AGE_SECS` | `86400` | `Cache-Control` max-age of the thumbnail 307 |

## Cutover runbook

Two switches must flip — the binary signs URLs, and Bunny enforces tokens — and they can't flip atomically. Browsers cache the 307 redirects (thumbnails 24h), so redirects minted before the cutover replay from caches after it. The phases below shrink that breakage from "up to 24h of broken thumbnails" to roughly a minute. One binary deploy total; everything else is env changes + restarts.

1. **Deploy this build in compatibility mode.** `CDN_TOKEN_AUTH=false`, `PLAYLIST_REDIRECT_MAX_AGE_SECS=60`, `THUMBNAIL_REDIRECT_MAX_AGE_SECS=60`. Behavior is identical to today except redirects expire from caches in a minute. Zero user impact.
2. **Wait ~24h** for the old long-lived cached redirects to drain.
3. **Flip both switches, code side first.** Restart with `CDN_TOKEN_AUTH=true` (+ `BUNNY_TOKEN_KEY` set, max-ages still `60`), then immediately enable token auth on the Bunny library (Stream → Library → Security). This order self-heals: signed URLs minted before Bunny flips 404 briefly, then become valid the instant the zone flips. Do **not** restore the long max-ages yet.
4. **Restore defaults.** Unset the two max-age overrides and restart.

Total user-visible breakage: the seconds-to-minutes gap in step 3, plus ≤60s of stale cached redirects. Newly uploaded videos are only exposed to the step-3 gap.

## Follow-up

- Remove the `CDN_TOKEN_AUTH` toggle once the cutover is complete — it's scaffolding, and a misconfigured environment could otherwise run unsigned indefinitely (startup warning aside).


rudyfraser (commented):
.

rudyfraser (approved):

PR #206, opened 2026-07-16
fix(common): from_micros_to_utc panics on real timestamps, 500ing refreshSession

rabble:
## Problem

`rsky_common::time::from_micros_to_utc` passes its **microsecond** argument to `NaiveDateTime::from_timestamp`, which interprets the value as **seconds**:

```rust
pub fn from_micros_to_utc(micros: i64) -> DateTime<UtcOffset> {
    let nanoseconds = 230 * 1000000;
    DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(micros, nanoseconds), Utc)
}
```

Any present-day timestamp (~`1.7e15` µs) is far outside chrono's representable second range, so the call panics:

```
panicked at chrono-0.4.41/src/lib.rs:709: invalid or out-of-range datetime
```

`AccountManager::rotate_refresh_token` formats the rotated expiry through `from_micros_to_str` → `from_micros_to_utc`, so **`com.atproto.server.refreshSession` returns a 500 on every call**:

```
>> Handler create_session panicked.
>> Outcome: Error(500 Internal Server Error)
```

The existing tests never caught it because the only in-tree callers pass tiny literals (`from_micros_to_str(1_000_000)`), and `1_000_000` seconds is a valid (if 1970-era) datetime.

## Fix

Use `DateTime::from_timestamp_micros`, which interprets the argument as microseconds and returns an `Option` rather than panicking internally. This also drops the fixed `230ms` of sub-second noise the old code stamped onto every converted instant.

## Tests

Added two regression tests in the existing `time` module. Both **panic with `invalid or out-of-range datetime` on the previous implementation** and pass on the fix:
- `from_micros_to_utc_handles_a_current_era_timestamp` — a 2023 timestamp round-trips and formats correctly.
- `from_micros_to_utc_preserves_sub_second_precision` — microsecond precision survives the round-trip.

Verified the whole `account_manager` suite (23 tests, incl. the token-rotation / expiry paths that consume `from_micros_to_str`) still passes — the fix preserves the "distant past = expired" semantics those tests rely on.

Scope is deliberately minimal: one function body plus tests, no signature or API changes.

### DIFF
--- a/rsky-common/src/time.rs
+++ b/rsky-common/src/time.rs
@@ -48,10 +48,16 @@ pub fn from_str_to_utc(str: &str) -> Result<DateTime<UtcOffset>> {
         .map_err(|e| anyhow!("failed to parse datetime {:?}: {}", str, e))
 }
 
-#[allow(deprecated)]
 pub fn from_micros_to_utc(micros: i64) -> DateTime<UtcOffset> {
-    let nanoseconds = 230 * 1000000;
-    DateTime::<Utc>::from_utc(NaiveDateTime::from_timestamp(micros, nanoseconds), Utc)
+    // NaiveDateTime::from_timestamp interprets its argument as SECONDS, so
+    // passing a microsecond value overflowed chrono's representable range and
+    // panicked ("invalid or out-of-range datetime"). rotate_refresh_token
+    // formats the rotated expiry through from_micros_to_str, so this panic
+    // surfaced as a 500 on every com.atproto.server.refreshSession call.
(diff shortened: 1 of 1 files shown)


rudyfraser (approved):
Verified against main: from_micros_to_utc passes microseconds to NaiveDateTime::from_timestamp (which expects seconds), panicking on any present-day timestamp and 500ing refreshSession via rotate_refresh_token. from_timestamp_micros is the correct fix; regression tests fail-on-old/pass-on-new. Approving.

PR #208, opened 2026-07-16
fix(pds): return ExpiredToken on an expired token, not InvalidRequest

rabble:
## Problem

An expired access or refresh token is reported as `400 {"error":"InvalidRequest","message":"BadJwt: \`Token has expired\`"}` — indistinguishable from a malformed token. A client can't tell that it should call `refreshSession`. The reference Bluesky PDS returns `400 {"error":"ExpiredToken","message":"Token is expired"}`.

## Fix

`ApiError::ExpiredToken` already exists and already renders the correct response — but it was only reachable from the refresh-rotation path, never from JWT expiry, because two layers erased the signal:

1. `access_check` collapsed every verify error into `AuthError::BadJwt`. This adds an `AuthError::ExpiredToken` variant and an `is_expired_jwt` helper that downcasts to `jwt_simple::JWTError::TokenHasExpired` (jwt-simple's `Error` *is* `anyhow::Error`, so the concrete enum is the real cause).
2. The guard sites hardcoded `ApiError::InvalidRequest` when populating the render cache. This adds `impl From<&AuthError> for ApiError` — `ExpiredToken → ExpiredToken`, **every other variant preserves the existing `InvalidRequest` rendering** — and routes the sites through it. The `Refresh` and `RevokeRefreshToken` guards, which bypass `access_check`, map expiry directly, so an expired *refresh* token on `refreshSession` also yields `ExpiredToken`.

Scope is deliberately minimal: only the expiry code path changes; all other auth errors render exactly as before.

## Tests

New integration tests (SQLite harness): expired access → `ExpiredToken`, expired refresh on `refreshSession` → `ExpiredToken`, malformed → still `InvalidRequest`, valid → 200. Tokens are backdated an hour, past jwt-simple's 900s tolerance. Verified the expiry tests fail on `main` for the right reason (`InvalidRequest` / "Token has expired") and pass with the change; full `rsky-pds` suite (259 lib + integration + oauth + space) stays green; clippy clean.

### DIFF
--- a/rsky-pds/src/apis/mod.rs
+++ b/rsky-pds/src/apis/mod.rs
@@ -1,4 +1,4 @@
-use crate::auth_verifier::{AccessStandard, AuthScope, Credentials};
+use crate::auth_verifier::{AccessStandard, AuthError, AuthScope, Credentials};
 use crate::handle;
 use crate::handle::errors::ErrorKind;
 use crate::pipethrough::{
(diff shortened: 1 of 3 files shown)


rudyfraser (approved):
Verified: verify_jwt propagates jwt-simple's JWTError undecorated through validate_bearer_token/validate_access_token (bare ? at every hop), so downcast_ref recovers TokenHasExpired; ExpiredToken renders 400 correctly; match arm ordering and From<&AuthError> preserve all non-expiry behavior. build-and-test green. (coverage-pds red is a pre-existing crate-wide gate issue, not this PR.)

PR #218, opened 2026-07-31
fix(rsky-video): re-encode uploads so Bluesky's federated ingest accepts them

rishibalakrishnan:
## Problem

Videos uploaded through the Blacksky app intermittently never play on bsky.app: the post renders but `video.bsky.app` returns a permanent 404 for the playlist, with no retry and no way to re-trigger ingest. The same content uploaded through Bluesky's own composer plays everywhere.

Root cause (established by controlled experiment on 2026-07-31, same account/content/minutes apart): Bluesky's federated video ingest deterministically rejects H.264 streams with **irregular frame timestamps** — the runt first-frame interval that copy-mode edits produce (trim / frame removal in Apple tooling, Twitter rips). rsky-video passed MP4 input through to the PDS untouched and stream-copied QuickTime, shipping those timestamps verbatim. Bluesky's own video service re-encodes every upload, which is why their uploads never hit this.

Experiment summary (all posted on a test account against production infra):
- original bytes, re-posted fresh → 404
- lossless remux (`-c copy`, new container) → 404
- re-encode at same 60 fps → **ingested in <60 s**
- re-encode at 30 fps → **ingested**

## Fix

Replace the MP4 passthrough and `-c copy` remux with a single normalization re-encode for every non-GIF upload (`reencode_to_mp4`): H.264 Main / yuv420p, source frame rate preserved, AAC 128k audio, faststart, even-dimension scale, video track required. Verified live: the previously-failing file, encoded with these exact settings and posted, ingested on `video.bsky.app` in ≤30 seconds.

Second commit hardens the encode path (review findings):
- **Still images rejected**: ffmpeg exposes PNG/JPEG as a 1-frame video stream, so `-map 0:v:0` alone would mint a broken one-frame video/mp4; the output frame count must be ≥2.
- **Post-encode size re-check**: the existing gate measures the input, but crf-23 output can exceed an already-tight source and `app.bsky.embed.video`'s maxSize; fail at the point of inflation instead of after PDS+Bunny work.
- **Bounded encoding**: 2-slot semaphore (x264 is thread-per-core; unbounded concurrent encodes starve the box) and a 10-minute kill-on-drop timeout for hung/adversarial input.

## Behavior changes

- webm/mkv input now converts to MP4 instead of failing at the PDS mimeType check.
- Every upload pays one x264 `veryfast` encode (this matches what Bluesky's service does). CPU/memory bounds beyond the semaphore remain with `harden/rsky-video-transcode-limits`.
- Audio is re-encoded to AAC 128k (also removes copy-mode failures for PCM/AMR audio).

## Tests

- `cargo test -p rsky-video` — unit tests pass.
- `cargo test -p rsky-video -- --ignored transcode` — 4 e2e tests (ffmpeg/ffprobe on PATH): timestamp normalization against a synthesized runt-interval poison fixture, container normalization (mov/m4v/3gp → isom), audio-only rejection, still-image rejection.

Known follow-ups (non-blocking, tracked in the design doc): runtime ffprobe assert of uniform output timestamps, quota charged on input rather than stored size, HDR→SDR tonemap, encode knobs → AppConfig.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -140,6 +140,16 @@ jobs:
         run: cargo build --release -p ${{ matrix.package }}
       - name: Run cargo test for ${{ matrix.package }}
         run: cargo test --release -p ${{ matrix.package }}
+      # The transcode e2e tests are #[ignore]d so plain `cargo test` doesn't
+      # require ffmpeg on every dev machine; CI installs it and runs them
+      # explicitly, otherwise the video normalization fix has zero coverage
+      # in the green build.
+      - name: Install ffmpeg for rsky-video e2e tests
+        if: matrix.package == 'rsky-video'
+        run: sudo apt-get update && sudo apt-get install -y ffmpeg
+      - name: Run rsky-video transcode e2e tests
+        if: matrix.package == 'rsky-video'
+        run: cargo test --release -p rsky-video -- --ignored transcode
 
   # Coverage gate for the sqlite PDS stack. Standalone so a coverage
   # failure never blocks CI for unrelated crates.
--- a/rsky-video/src/main.rs
+++ b/rsky-video/src/main.rs
@@ -65,6 +65,13 @@ async fn main() -> color_eyre::Result<()> {
         config.host, config.port
     );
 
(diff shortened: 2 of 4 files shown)


afbase (approved):
+1

PR #232, opened 2026-08-16
fix(pds): space reads before the first write are empty, not missing

ngerakines:
### The problem

`SpaceStore::live_repo_state` answers two questions at once — is this repo tombstoned, and does a `space_repo` row exist — and six space read handlers call it as a precondition. The row is created by the member's **first write** in `apply_writes_tx`, so a member who has been added to a space and has not written yet gets `400 SpaceNotFound` from every read of their repo.

Every member passes through that window exactly once, and an app that syncs a member's repo on page load meets them in it. Observed against a live space whose authority runs a different implementation: `addMember` succeeded, and every subsequent `com.atproto.space.listRecords` for that member returned `SpaceNotFound` until they wrote — which they could not usefully do while the app could not read them back.

`simplespace/add_member.rs` already names the ordering:

> A membership must be discoverable through the member's own listSpaces **before their first write creates a repo row**.

### The change

A new `SpaceStore::readable_repo_state` answers only what a read needs: `Ok(None)` before the first write, and the same `SpaceDeleted` error for a tombstoned repo. Five handlers swap to it and fall through to their naturally empty results; `list_repo_ops` returns an empty page so an incremental syncer can start from `since = None`.

| | before | after |
|---|---|---|
| `listRecords` | `400 SpaceNotFound` | `200 { records: [] }` |
| `getRecord` | `400 SpaceNotFound` | `RecordNotFound` |
| `listBlobs` | `400 SpaceNotFound` | empty page |
| `getBlob` | `400 SpaceNotFound` | `BlobNotFound` |
| `registerNotify` | `400 SpaceNotFound` | registers |
| `listRepoOps` | `400 SpaceNotFound` | empty page |
| tombstoned repo | `SpaceDeleted` | `SpaceDeleted` (unchanged) |

**Deliberately unchanged:** `getRepo` and `getLatestCommit` both sign a commit from `state.rev`/`state.hash()`, and an unwritten repo has no commit to sign. Erroring there is defensible — it just should not be the same error that says the space is missing. Happy to follow up separately if you'd like a distinct code.

**One existing assertion changed:** `repo_lifecycle_flags_and_listing` pinned the old `list_repo_ops` behaviour. That is the one behavioural change worth a second look.

The write path is untouched — `applyWrites` and `putRecord` never required the row and already create it on first commit.

### Verification

- `cargo test -p rsky-pds --lib` — 286 passed
- `cargo fmt --check`, `cargo clippy -p rsky-pds --lib` clean
- New `reads_before_the_first_write_are_empty_not_missing` fails without the fix (verified by reverting `readable_repo_state` to delegate to `live_repo_state`)
- New `readable_repo_state_still_refuses_a_deleted_repo` covers the tombstone half

### Suggested conformance check

*An invited member with no records reads as empty.* Every implementation passes through that state, and it is easy to never test — anything that writes before it reads skips the window entirely.

### DIFF
--- a/rsky-pds/src/actor_store/space/mod.rs
+++ b/rsky-pds/src/actor_store/space/mod.rs
@@ -364,6 +364,25 @@ impl SpaceStore {
         Ok(state)
     }
 
+    /// Repo state for a *read*: `Ok(None)` before the member's first write.
+    ///
+    /// The `space_repo` row is created by [`apply_writes_tx`], so a member who
+    /// has been added to a space but has not written to it yet has no row. That
+    /// is the ordinary pre-first-write state rather than a missing space — the
+    /// same ordering `simplespace::add_member` documents when it indexes a
+    /// membership "before their first write creates a repo row".
+    ///
+    /// A tombstoned repo still errors, which is the half of
+    /// [`Self::live_repo_state`] a read actually needs.
+    pub async fn readable_repo_state(&self, space_uri: &str) -> Result<Option<SpaceRepoState>> {
+        match self.repo_state(space_uri).await? {
+            Some(state) if state.deleted => {
+                Err(SpaceStoreError::SpaceDeleted(space_uri.to_string()).into())
+            }
+            other => Ok(other),
+        }
+    }
+
     pub async fn get_record(
         &self,
         space_uri: &str,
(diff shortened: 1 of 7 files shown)


rudyfraser:
The `listRecords` change is correct, but `listRepoOps` still returns `SpaceNotFound` before the first write.

`SpaceStore::list_repo_ops` now returns `([], false)` for a missing `space_repo` row. The handler then sees `has_more == false` and calls `serve_commit`, which calls `live_repo_state` and fails on the same missing row.

For an admitted member with no commits, return `200` with an empty `ops` array and no `commit`. Keep the signed `commit` for existing, caught-up repos. Please add an HTTP-level regression test covering addMember -> no write -> `listRecords` and `listRepoOps`.

PR #239, opened 2026-08-24
fix(wintermute): suppress duplicate_table NOTICE flood from bulk staging tables

afbase:
The bulk indexer creates its `_bulk_*` staging tables with `CREATE TEMP TABLE IF NOT
EXISTS` on every batch. Pools recycle with `RecyclingMethod::Fast`, which never discards
temp tables, so they persist for the life of a connection and every batch after the first
raises a `duplicate_table` NOTICE that tokio-postgres logs at INFO. That stream dominates
the log output and shortens how far back the journal reaches, which makes incident
forensics harder.

This adds `config::pg_connect_options()` and sets it on every pool, so the server is told
`client_min_messages=warning` and never generates or transmits the notice. A client-side
log filter would still pay for generating and sending it.

The redundant DDL itself is left in place here and addressed separately, because removing
it requires every pool that reaches the bulk path to pre-create the staging tables.

## Test plan

- [ ] `cargo clippy -p rsky-wintermute --all-targets --no-deps -- -D warnings`
- [ ] `cargo test -p rsky-wintermute`
- [ ] `cargo build --release -p rsky-wintermute`
- [ ] Indexer logs show no `tokio_postgres` NOTICE lines after restart
- [ ] Indexing throughput and error rates unchanged
- [ ] `direct_index` and `car_loader` still index a repo end to end

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8512,7 +8512,7 @@ dependencies = [
 
 [[package]]
 name = "rsky-wintermute"
-version = "0.9.2"
+version = "0.9.3"
 dependencies = [
  "base64 0.22.1",
  "bytes",
--- a/rsky-wintermute/Cargo.toml
+++ b/rsky-wintermute/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rsky-wintermute"
-version = "0.9.2"
+version = "0.9.3"
 edition = "2024"
 authors = ["Rudy Fraser <him@rudyfraser.com>"]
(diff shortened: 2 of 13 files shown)


afbase:
Superseded by #240, which now targets `main` and includes this commit unchanged. The two changes are not independently mergeable — `config::create_pg_pool` in #240 calls `pg_connect_options()` added here — so they are better reviewed as one PR with two commits than as a stack.

Branch `fix/wintermute-suppress-notice-flood` is left in place; #240 builds on it.

PR #246, opened 2026-09-02
fix(rsky-crypto): guard has_prefix against inputs shorter than the prefix

afbase:
rsky-crypto parses `did:key` identifiers by scanning a multicodec plugin table over decoded multibase bytes. `has_prefix` compared that slice without checking its length first, so a `did:key` whose payload was shorter than a two-byte prefix panicked rather than being rejected as an unsupported key type. This adds the length guard and covers the parse and verify entrypoints.

Test plan:
- [ ] `cargo test -p rsky-crypto` passes
- [ ] the three new tests fail when the length guard is reverted
- [ ] `cargo build --release -p rsky-crypto` succeeds
- [ ] coverage for `rsky-crypto` does not regress

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2464,6 +2464,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "dhat"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


afbase (commented):
LGTM

PR #247, opened 2026-09-02
fix(rsky-repo): verify commit signatures with digest semantics on both curves

afbase:
rsky-crypto exposes two verification entrypoints whose `data` argument means different things per curve: `verify_signature` treats it as a digest on secp256k1 but hashes it internally on p256, while `verify_signature_digest` gives both curves digest semantics. `verify_commit_sig` computes a sha256 digest and called the former, so ES256 commit signatures were checked against a double hash and never verified. It failed closed, returning `Ok(false)` rather than an error, so it read as a bad signature. The secp256k1 path is unchanged.

Test plan:
- [ ] `cargo test -p rsky-repo` passes, including the existing secp256k1 commit tests
- [ ] the new p256 test fails when the call is reverted to `verify_signature`
- [ ] `cargo build --release -p rsky-repo` succeeds

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8339,7 +8339,7 @@ dependencies = [
 
 [[package]]
 name = "rsky-repo"
-version = "0.0.6"
+version = "0.0.7"
 dependencies = [
  "anyhow",
  "async-recursion",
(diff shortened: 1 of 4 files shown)


afbase:
needs review.  will assess later

PR #248, opened 2026-09-02
feat(rsky-pds)!: give every account its own repo signing key

afbase:
Every actor store was built with the process-wide repo signing key, so one leaked secret could forge commits — and mint service-auth tokens — for any account on the server. There was no per-account blast radius.

`createAccount` now generates a secp256k1 keypair per account and stamps it into that account's PLC operation. `getRecommendedDidCredentials`, `submitPlcOperation` and `activateAccount` validate against the account's own key, and service-auth JWTs are signed with the issuing account's key. Those change together by necessity: once a DID document names an account's own key, a token signed with the global key no longer verifies.

Accounts created before this release keep the shared key until an operator runs the new `rotate-keys` binary, which publishes a PLC operation and emits firehose events per account and so is never run automatically. It is resumable and idempotent — an account whose document already names its own key is skipped, and one whose key was written but whose operation did not land is republished rather than re-keyed. Operator steps and the warnings that go with them are in `rsky-pds/MIGRATING-1.0.md`.

Breaking, hence the major version. `PDS_REPO_SIGNING_KEY_K256_PRIVATE_KEY_HEX` remains required: `rotate-keys` uses it to tell a rotated account from an unrotated one.

Test plan:
- [ ] `cargo test -p rsky-pds` passes
- [ ] `cargo build --release -p rsky-pds` succeeds and produces `rotate-keys`
- [ ] two newly created accounts have different signing keys, and neither matches the configured shared key
- [ ] an inbound account migration completes end to end against a real relay and appview
- [ ] `rotate-keys --dry-run` reports the unrotated accounts and changes nothing
- [ ] `rotate-keys` on a test account: repo still verifies downstream, `getRecommendedDidCredentials` returns that account's key
- [ ] a service-auth token verifies against the issuing account's DID document key
- [ ] re-running `rotate-keys` after a completed run is a no-op

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8211,7 +8211,7 @@ dependencies = [
 
 [[package]]
 name = "rsky-pds"
-version = "0.13.18"
+version = "1.0.0"
 dependencies = [
  "anyhow",
  "argon2",
--- a/rsky-pds/Cargo.toml
+++ b/rsky-pds/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rsky-pds"
-version = "0.13.18"
+version = "1.0.0"
 authors = ["Rudy Fraser <him@rudyfraser.com>"]
 description = "Rust reference implementation of an atproto PDS."
 license = "Apache-2.0"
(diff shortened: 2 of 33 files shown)


afbase (commented):
lgtm

PR #250, opened 2026-09-03
fix(pds): switch password hashing from Argon2 to scrypt, matching upstream

afbase:
## Summary
- rsky-pds hashed passwords with Argon2; the reference TypeScript PDS uses scrypt, so accounts/rows are not portable between the two implementations without forcing a mass password reset.
- Switches `hash()` to scrypt using parameters and stored-hash encoding matched byte-for-byte to `atproto/packages/pds/src/account-manager/helpers/scrypt.ts` (N=16384, r=8, p=1, 64-byte key, `<salt>:<hex key>` format) — including a subtle detail: Node's `crypto.scrypt` hashes the salt as its literal ASCII string rather than hex-decoding it first, which is now replicated exactly and was cross-verified against independent Python/Rust reference generators.
- Also fixes app-password salting (`sha256(did)[:16]`, hex) to match upstream, which previously used a different scheme.
- `verify()` dispatches by encoding so existing Argon2-hashed rows (`$`-prefixed PHC strings) keep verifying forever — no lockout for currently deployed accounts — while all new hashes are scrypt going forward.
- Checked cocoon and tranquil-pds: both use bcrypt, not scrypt/Argon2, so neither adds an additional interop constraint.

## Test plan
- [x] `cargo test -p rsky-pds` — full suite green (308+ tests), including 5 new tests: round-trip, hand-constructed pds-format interop vector, legacy-Argon2-still-verifies, malformed-hash rejection, arbitrary-salt permissiveness

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8266,7 +8266,7 @@ dependencies = [
 
 [[package]]
 name = "rsky-pds"
-version = "1.0.0"
+version = "1.1.0"
 dependencies = [
  "anyhow",
  "argon2",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -662,7 +662,7 @@ dependencies = [
  "bytes",
  "fastrand",
  "hex",
- "hmac",
+ "hmac 0.12.1",
  "http 0.2.12",
  "http 1.3.1",
  "http-body 0.4.6",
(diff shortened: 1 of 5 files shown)


afbase (commented):
LGTM

afbase (commented):
LGTM

PR #251, opened 2026-09-03
feat(pds): enforce blob/rpc/identity/account OAuth scopes (proposal 0011)

afbase:
Proposal 0011 defines five OAuth resource types: `repo`, `rpc`, `blob`, `identity`, `account`. rsky-pds parsed three and enforced one, so a granular OAuth token bought app-password-equivalent access outside record writes.

- Adds `Identity`/`Account` variants to the scope grammar, parsed with the same shorthand/query-form rules `repo:` already has.
- Expresses enforcement as request guards, so a handler declares its requirement in its signature rather than calling a check inline.
- Generalizes permission-set expansion: a set's `repo`/`rpc`/`blob`/`identity`/`account` entries previously expanded to nothing, and the resolved scopes are now fed through the same parser an inline scope goes through.
- **Gates scope checks on auth source rather than scope content.** The helpers previously required the session to already hold a grant of a resource before checking that resource, so a session granted only `repo:` had no `Blob`/`Rpc`/`Identity`/`Account` variants, every gate was false, and it reached unrestricted access on all four. A session with no granted scopes is exempt, `transition:generic` is exempt where the reference implementation says so, and every other OAuth session must produce a matching grant.
- Fixes a second fail-open where an audience `assert_rpc_scope` could not resolve returned `Ok(())` and proxied the call anyway.

| resource | parsed (before → after) | enforced (before → after) |
|---|---|---|
| repo | yes | yes (gate corrected) |
| blob | yes | no → **yes** |
| rpc | yes | no → **yes** |
| identity | no variant → yes | no → **yes** |
| account | no variant → yes | no → **yes** |

## Behavior change

Absence of a grant is now a denial. This affects granular OAuth sessions only — app passwords and legacy access tokens carry no granted scopes and are unaffected.

- A granular session must hold an explicit `rpc:<lxm>?aud=<did>` grant to use appview proxying. A client that requested `repo:`/`blob:` but no `rpc:` loses proxying it previously had. `transition:generic` sessions are exempt, so this does not affect clients using the legacy transition scopes.
- A session holding only `space:` grants is now denied `com.atproto.repo.*` writes and proxying.
- A deployment whose appview URL/DID is unset, or whose proxy target fails the SSRF check, now refuses granular-session proxying rather than allowing it. Worth a deploy-time config check.
- Permission-set resolution becomes load-bearing for access rather than only for narrowing: an unresolvable set now denies rather than silently widening, so a resolution outage becomes visible to clients.

## Transitional scopes, per the spec

The [OAuth spec](https://atproto.com/specs/oauth#transitional-scopes) defines `transition:generic` as repo writes, blob uploads, preferences, and service proxying — and explicitly **"no account management actions: change handle, change email, delete or deactivate account, migrate account"**. This PR now matches that exactly:

| | `transition:generic` |
|---|---|
| `repo:` write | permitted |
| `blob:` upload | permitted |
| `rpc:` proxying | permitted |
| `identity:` change handle | **denied** |
| `account:` change email / deactivate / migrate | **denied** |

`transition:email` is unaffected — it confers a read on the address, and every account guard here asks for `manage`.

The user-visible consequence: clients holding `atproto transition:generic transition:email transition:chat.bsky` now get 403 on `updateHandle`, `updateEmail`, `deactivateAccount` and the PLC operations, and must request the corresponding `identity:` / `account:` scope. That is the spec'"'"'s intent, but it is the change most likely to be noticed.

Unrelated and pre-existing, noted while checking the above: `com.atproto.server.getSession` returns the account email unconditionally, gating on no scope at all. Worth its own issue.

## Comparison with other implementations

Corrected after reading each codebase directly; an earlier draft of this description got the tranquil-pds claims wrong.

- **TypeScript reference** — `authorize` is a required field on the auth-verifier options type, so a route cannot be registered without declaring what it permits, and there are 27 explicit opt-out sites of two legible kinds. Permission-set expansion happens at token issuance and an unresolvable set throws rather than returning empty.
- **tranquil-pds** — enforces all five at handlers (identity at 4, account at 7, rpc at 2), with a typestate proof that is compile-time enforced for repo writes. Its gate keys on auth source rather than scope content, which is the shape adopted here; it has no fail-open equivalent, and an empty expansion is a hard error that refuses the token.
- **cocoon** — parses seven resources and fails closed on unparseable scopes at PAR/authorize, but enforces only `repo` and `blob`, and `importRepo` carries no scope check at all.

Enforcement here remains opt-in per route: a new route can take a plain auth guard and get no check. Making the declaration mandatory is tracked separately.

## Test plan

- [x] `cargo fmt` clean (two pre-existing violations on main in `password.rs` and `get_service_auth.rs` are untouched)
- [x] `cargo clippy -p rsky-pds --all-targets --no-deps -- -D warnings` clean
- [x] `cargo test -p rsky-pds` — 386 passed, 0 failed, 1 pre-existing ignored (incl. a case pinning the full transition:generic surface)
- [x] Three tests that asserted the fail-open as intended are inverted
- [x] New: a `repo:`-only grant is denied the four resources it does not name; an empty scope set, an `include:` resolving to nothing, and an unparseable scope are each denied all five; `transition:generic` keeps its legacy reach; a session without granted scopes is unaffected
- [x] New end-to-end through the full Rocket/DPoP stack: a `repo:`-only session gets 403 on uploadBlob, updateHandle and submitPlcOperation; an unresolvable proxy audience refuses instead of proxying
- [ ] Confirm against a real granular OAuth client that the `rpc:` requirement does not strand it
- [ ] Confirm the transition-scope surface above against a real client before deploy

### DIFF
--- a/rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
+++ b/rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
@@ -22,6 +22,11 @@ pub async fn sign_plc_operation(
     auth: AccessFull,
     account_manager: AccountManager,
 ) -> Result<Json<Operation>, ApiError> {
+    crate::apis::assert_account_scope(
+        &auth.access.credentials,
+        "repo",
+        crate::oauth_scope::AccountAction::Manage,
+    )?;
     let did = auth.access.credentials.unwrap().did.unwrap();
     let request = body.into_inner();
     let token = request.token.clone();
--- a/rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
+++ b/rsky-pds/src/apis/com/atproto/identity/sign_plc_operation.rs
@@ -1,7 +1,7 @@
 use crate::account_manager::AccountManager;
 use crate::apis::com::atproto::server::PDS_PLC_ROTATION_KEYPAIR;
 use crate::apis::ApiError;
-use crate::auth_verifier::AccessFull;
+use crate::auth_verifier::AccountRepoScopedAccess;
 use crate::models::models::EmailTokenPurpose;
 use crate::plc;
 use crate::plc::operations::create_update_op;
@@ -19,14 +19,10 @@ use std::collections::BTreeMap;
 #[tracing::instrument(skip_all)]
 pub async fn sign_plc_operation(
     body: Json<SignPlcOperationRequest>,
-    auth: AccessFull,
+    // `AccessFull` (its pre-existing tier) via the guard's default `Base`.
+    auth: AccountRepoScopedAccess,
     account_manager: AccountManager,
 ) -> Result<Json<Operation>, ApiError> {
-    crate::apis::assert_account_scope(
-        &auth.access.credentials,
-        "repo",
-        crate::oauth_scope::AccountAction::Manage,
-    )?;
     let did = auth.access.credentials.unwrap().did.unwrap();
     let request = body.into_inner();
     let token = request.token.clone();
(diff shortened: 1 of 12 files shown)


afbase (commented):
LGTM

PR #252, opened 2026-09-03
fix(pds): correct inverted service-auth privilege check

afbase:
## Summary
`com.atproto.server.getServiceAuth` had its privilege check backwards: rsky denied a service-auth token when the caller **was** privileged, and allowed it otherwise. The TS reference denies when the requested method is privileged **and the caller is not**. In practice this meant a plain app-password session could mint a service-auth token for privileged `chat.bsky.*` methods (read DMs, send messages, list conversations, export account data), while legitimately privileged clients were incorrectly rejected.

- Fixes the condition to match upstream exactly: deny iff `PRIVILEGED_METHODS.contains(lxm) && !is_privileged`.
- Adds `aud` DID-format validation on the same endpoint (cocoon independently validates this too via a struct tag, confirming it's an established pattern worth having).
- **Found and fixed a second, more serious bug surfaced by testing the first fix end-to-end**: `is_privileged` was hardcoded to `None` in the real request path (`validate_access_token`/`validate_dpop_access_token` in `auth_verifier.rs`, which is what `AccessFull`/`AccessPrivileged` actually use) — so in production the corrected condition alone would have denied **every** `chat.bsky.*` service-auth request, including from legitimately privileged sessions. Extracted a shared `AuthScope::is_privileged()` helper and applied it at all three call sites that build `Credentials`.
- Compared against cocoon and tranquil-pds: cocoon has no `PRIVILEGED_METHODS` concept at all (mints for any lxm short of recursive tokens); tranquil-pds uses a positive-grant scope check that can't produce this class of inversion by construction, and also implements the takedown carve-out directly in its `get_service_auth` handler — confirming that carve-out belongs at the endpoint, which rsky does not yet do (flagged, not fixed here — rsky's `AccessFull` guard doesn't currently enforce takedown status at all, so adding the carve-out first needs the underlying check plumbed in, which is out of scope for this fix).

## Test plan
- [x] Unit tests on the extracted condition, verified to fail against the original inverted logic and pass against the fix
- [x] Two new end-to-end Rocket integration tests in `tests/signing_key_tests.rs` (real login → bearer token → `GET /xrpc/com.atproto.server.getServiceAuth`): a full-access session is allowed for a privileged method; an app-password session is denied. Verified the first test actually catches the `is_privileged`-always-`None` bug by reverting it locally and watching the test fail (500 instead of 200) before re-applying the fix.
- [x] `cargo build -p rsky-pds` clean, `cargo test -p rsky-pds` — all suites green (310+ lib tests, `signing_key_tests` 5/5)

### DIFF
--- a/rsky-pds/src/apis/com/atproto/server/get_service_auth.rs
+++ b/rsky-pds/src/apis/com/atproto/server/get_service_auth.rs
@@ -10,7 +10,39 @@ use rocket::serde::json::Json;
 use rocket::State;
 use rsky_common::time::{from_micros_to_utc, HOUR, MINUTE};
 use rsky_lexicon::com::atproto::server::GetServiceAuthOutput;
+use rsky_syntax::did::ensure_valid_did;
 use std::time::SystemTime;
+
+/// Denies a request for a service-auth token when the *requested* method
+/// (`lxm`) is privileged (the `chat.bsky.*` surface plus
+/// `com.atproto.server.createAccount`) and the *caller's own* session is not
+/// itself privileged.
+///
+/// Mirrors the upstream TS reference (`getServiceAuth.ts`):
+/// `lxm != null && PRIVILEGED_METHODS.has(lxm) && !isAccessPrivileged(scope)`.
+///
+/// This is intentionally the inverse of a naive "gate privileged sessions"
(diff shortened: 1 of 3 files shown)


afbase (commented):
lgtm

PR #253, opened 2026-09-03
fix(pds): describe OAuth scopes in plain language, redesign consent UI

afbase:
## Summary
- Fixes a real bug on the OAuth consent screen: it rendered each granted scope's raw wire token near-verbatim. `scope_description()` only recognized four literal legacy strings and fell back to a generic "additional access requested by the app" blurb for every modern scope form (`repo:`, `blob:`, `rpc:`, `include:`, `space:`) — so a user approving e.g. `include:app.bsky.authFull` never saw what it actually granted.
- `describe_scope()` now parses every resolvable scope form through the crate's own `oauth_scope`/`space_scope` modules (reusing the real enforcement parser, so the description can never drift from what's actually enforced) into a plain-language title + detail line.
- `include:` permission sets still can't be expanded at render time — that needs an async DNS+HTTP fetch this synchronous per-request template render has no path to. Left a narrow, honest `TODO` at the exact spot that wiring belongs, with user-facing copy that says so rather than guessing.
- Redesigned the three OAuth templates (consent/signin/error) and shared stylesheet: real type scale, card layout, scope rows as distinct chips with a granted-permission indicator, light/dark palette tokens borrowed from blacksky.community's product palette (`src/alf/themes.ts`). Deliberately no brand marks/logo (this page ships to any rsky-pds deployment) and no webfont dependency (a security-sensitive page shouldn't add a third-party font-CDN dependency for arbitrary self-hosted deployments) — uses the same system-font fallback stack the product itself falls back to.

## Test plan
- [x] Added `examples/render_oauth_consent.rs`, rendering the real production `ConsentPage`/`SignInPage`/`ErrorPage` Askama templates with a representative 8-scope grant; screenshotted with headless Chrome in both light and dark to visually confirm the fix and layout
- [x] New test: `renders_modern_scope_grammar_as_plain_language`
- [x] `cargo test -p rsky-pds --lib` — 304 passed, 1 pre-existing ignored network test, 0 failed
- [x] `cargo fmt` / `cargo clippy -p rsky-pds --all-targets` clean


afbase (commented):
lgtm

PR #255, opened 2026-09-03
refactor(pds): make an OAuth scope declaration required to learn the requester

afbase:
Stacked on #251 — base that PR, not `main`, so this shows only its own diff. Merge #251 first.

Scope enforcement is opt-in. A route can take a plain access guard, read the requester's DID off the credentials, and perform a write with no scope check, and nothing detects the omission. That is the shape of defect that left `blob`, `rpc`, `identity` and `account` unenforced to begin with — and the same shape appears in other implementations, so it is a property of the design rather than an oversight.

`Scoped<D, Base>` makes the declaration structural: `D` names what the route must be permitted to do, and the requester's DID is reachable only from `Scoped`, only through an accessor that has already run `D`'s check. A route cannot learn whose repo it is writing without first saying what it needs.

Routes that require nothing say so explicitly, mirroring the reference implementation's two `authorize` opt-outs:

- `NoScopeRequired` — any session the route's tier admits may call it (`authorize: () => {}`)
- `OAuthForbidden` — OAuth cannot reach this route; app passwords and legacy tokens only (`authorize: () => { throw new ForbiddenError(...) }`)

Both are one grep away, so every unenforced route can be enumerated:

    rg "Scoped<(NoScopeRequired|OAuthForbidden)" rsky-pds/src/apis/

## Where to spend review attention

Most of the diff is mechanical. The substance is in two places:

- `rsky-pds/src/auth_verifier/scope.rs` — the `ScopeDecl` trait, `Scoped`, and the declarations
- the two `compile_fail` doctests on `Scoped`, which pin the actual property: a route holding a plain access guard cannot read the DID, and a deferred declaration cannot skip the value its check needs

Everything else is migrating ~35 call sites off manual `credentials.did` unwrapping onto `auth.did().await?`. The five hand-written scoped guards from #251 collapse into instances of the generic, which is why `auth_verifier.rs` gets substantially shorter.

## Not covered

Enforcement is now unforgettable for every route that goes through `Scoped`. Routes still on the older access guards are unchanged in behaviour and remain to be migrated; they are visible as the guards they still take. This PR does not attempt all 199 registrations.

## Test plan

- [x] `cargo fmt` clean (the two pre-existing violations on main in `password.rs` and `get_service_auth.rs` are untouched)
- [x] `cargo clippy -p rsky-pds --all-targets --no-deps -- -D warnings` clean
- [x] `cargo test -p rsky-pds` — 398 passed, 0 failed, 1 pre-existing ignored, plus 2 doctests
- [x] `compile_fail` doctests confirm the property holds at compile time — verified by hand that they fail for the right reasons (`E0616` private field; `E0599` unsatisfied trait bounds), by temporarily running them as ordinary doctests
- [ ] Consider a `trybuild` harness. Each block names its expected error, but rustdoc on stable checks only *that* compilation fails, not the code — swapping in a deliberately wrong code still passes. So a snippet that starts failing for an unrelated reason (a renamed import, a moved module) would go quietly hollow. `trybuild` compares full stderr and would hold the distinction, at the cost of one dev-dependency
- [ ] Confirm the `OAuthForbidden` routes are the right set — that list is a policy decision, not a mechanical one
- [ ] Decide whether the remaining unmigrated routes land here or as a follow-up

### DIFF
--- a/rsky-pds/src/apis/app/bsky/actor/get_preferences.rs
+++ b/rsky-pds/src/apis/app/bsky/actor/get_preferences.rs
@@ -1,7 +1,7 @@
 use crate::actor_store::blobstore::BlobstoreFactory;
 use crate::actor_store::ActorStore;
 use crate::apis::ApiError;
-use crate::auth_verifier::AccessStandard;
+use crate::auth_verifier::scope::{NoScopeRequired, Scoped};
 use anyhow::Result;
 use rocket::serde::json::Json;
 use rocket::State;
@@ -9,11 +9,11 @@ use rsky_lexicon::app::bsky::actor::{GetPreferencesOutput, RefPreferences};
 
 async fn inner_get_preferences(
     blobstore_factory: &State<BlobstoreFactory>,
-    auth: AccessStandard,
+    auth: Scoped<NoScopeRequired>,
     actor_store: &State<ActorStore>,
 ) -> Result<GetPreferencesOutput> {
-    let auth = auth.access.credentials.unwrap();
-    let requester = auth.did.unwrap().clone();
+    let credentials = auth.credentials().await?.clone().unwrap();
+    let requester = auth.did().await?;
     let actor_store = actor_store
         .read(
             requester.clone(),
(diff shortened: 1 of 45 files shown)


afbase (commented):
This is a really good start for proposal 11 scope types over a broad set of endpoints + the auth_verifier.rs that gives the right layer of abstraction to cover this _generally_ in the future.

PR #256, opened 2026-09-04
fix(pds): withhold the account email from sessions not granted it

afbase:
Stacked on #251 for the `account:` scope variants it adds — base that PR, not `main`. Sibling to #255; the two are independent.

`com.atproto.server.getSession` returned the email address and its confirmation state to every session, consulting no scope at all. Any OAuth token reached them, including one granted nothing but repo writes.

The [OAuth spec](https://atproto.com/specs/oauth#transitional-scopes) confers the address through `transition:email` — whose entire definition is that it *"gets included in response to `com.atproto.server.getSession`"* — or through an explicit `account:email` grant. If email were unconditional, that scope would grant nothing and have no reason to exist.

| session | email |
|---|---|
| app password / legacy `createSession` (no granted scopes) | returned |
| `transition:email` | returned |
| `account:email`, `?action=read`, `?action=manage` | returned |
| `transition:generic` | **withheld** |
| `repo:`-only, or any other granular grant | **withheld** |

This degrades the response rather than refusing the request — the endpoint still answers, minus the address. That is the `allows_*` half of the scope API rather than the `assert_*` half.

## Two details

`email` had no `skip_serializing_if`, so withholding it would have emitted `"email": null` where the lexicon types the field as a string. The lexicon marks only `handle` and `did` as required, so omission is the schema-valid way to withhold it; `emailConfirmed` already had the attribute. Hence the `rsky-lexicon` patch bump.

The handler also did `credentials.unwrap().did.unwrap()`. Those now return `AccountNotFound` instead of panicking.

## How other implementations handle this

Read directly rather than assumed:

- **TypeScript reference** — redacts unless `account:email?action=read`, stripping `email`, `emailConfirmed` and `emailAuthFactor`. Its route declares `authorize: () => {}` with the comment *"Always allowed. `email` access is checked in the handler."*
- **tranquil-pds** — redacts on `allows_email_read()`, which is `transition:email` OR an `account:email`/`account:*` grant. It computes the predicate without first asking whether a scope check applies, so legacy password sessions appear unable to see their own email; this PR avoids that by routing through the same no-granted-scopes exemption the other checks use.
- **cocoon** — returns the email unconditionally. The identical gap, in the same endpoint, worth reporting upstream.

rsky has no `emailAuthFactor` field in `GetSessionOutput`, so there are two fields to withhold here rather than upstream's three.

## Behavior change

A granular OAuth session that never requested `transition:email` or `account:email` stops seeing the address. Sessions with no granted scopes — the ordinary app-password path — are unaffected. Lands in the same deploy as #251's transition-scope changes and would be noticed alongside them.

## Test plan

- [x] `cargo fmt` clean
- [x] `cargo clippy -p rsky-pds --all-targets --no-deps -- -D warnings` clean (4 pre-existing `rsky-lexicon` lints are identical on the base)
- [x] `cargo test -p rsky-pds` — 387 passed, 0 failed, 1 pre-existing ignored
- [x] New unit test pinning the full visibility matrix above
- [ ] Confirm against a real OAuth client that losing the address is acceptable, or that the client can request `account:email`
- [ ] Consider whether `getSession` should advertise the requirement in its lexicon description

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8225,7 +8225,7 @@ dependencies = [
 
 [[package]]
 name = "rsky-lexicon"
-version = "0.10.2"
+version = "0.10.3"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
--- a/rsky-lexicon/Cargo.toml
+++ b/rsky-lexicon/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rsky-lexicon"
-version = "0.10.2"
+version = "0.10.3"
 edition = "2021"
 publish = true
 description = "Bluesky API library"
--- a/rsky-lexicon/src/com/atproto/server.rs
+++ b/rsky-lexicon/src/com/atproto/server.rs
@@ -171,6 +171,7 @@ pub struct CreateSessionOutput {
 pub struct GetSessionOutput {
     pub handle: String,
     pub did: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub email: Option<String>,
     #[serde(rename = "emailConfirmed", skip_serializing_if = "Option::is_none")]
     pub email_confirmed: Option<bool>,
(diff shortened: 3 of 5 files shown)


afbase (commented):
LGTM

PR #258, opened 2026-09-07
style(pds): apply cargo fmt to password helper tests

afbase:
## Summary

`cargo fmt -- --check` fails on `main`, so the workspace `formatting` CI job is red for every branch. This applies rustfmt to the one file that had drifted.

## Test plan

- [ ] `cargo fmt -- --check` exits 0
- [ ] `cargo check -p rsky-pds` succeeds


afbase (commented):
LGTM

PR #259, opened 2026-09-07
fix(repo): verify commit signatures against the digest for both curves

afbase:
## Summary

`verify_commit_sig` hashed the DAG-CBOR commit and then called `verify_signature`, whose per-curve plugins disagree on what `data` means: secp256k1 treats it as a digest, P-256 hashes it internally. P-256-keyed repos were therefore hashed twice and their commit signatures never verified. `verify_signature_digest` gives both curves digest semantics and is used instead.

Also vendors the atproto interop signature vectors (valid, high-S, and DER-encoded cases) and runs them against the verifier.

## Test plan

- [ ] `cargo test -p rsky-repo -p rsky-crypto` passes
- [ ] The interop vectors all match their expected verdict
- [ ] Reverting the one-line change fails the P-256 regression test and not the secp256k1 one
- [ ] `cargo build --release -p rsky-repo -p rsky-crypto` succeeds

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8143,9 +8143,12 @@ name = "rsky-crypto"
 version = "0.2.1"
 dependencies = [
  "anyhow",
+ "base64 0.22.1",
  "multibase",
  "p256 0.13.2",
  "secp256k1",
+ "serde",
+ "serde_json",
  "sha2 0.10.9",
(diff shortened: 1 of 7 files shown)


afbase (commented):
LGTM

PR #260, opened 2026-09-07
fix(relay): reject malleable high-S signatures

afbase:
## Summary

The relay validator verified commit and label signatures without a low-S check, so a malleable high-S signature over the same message was accepted. atproto requires low-S. Both curve paths are folded into one helper that rejects when `normalize_s` reports the signature was not already low-S.

Also vendors the atproto interop signature vectors and runs them against the validator, and serializes three tests that flip the process-global `SHUTDOWN` flag while others depend on it being clear.

## Test plan

- [ ] `cargo test -p rsky-relay` passes, repeatedly and under parallel load
- [ ] The interop vectors all match their expected verdict
- [ ] `cargo build --release -p rsky-relay` succeeds

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8439,7 +8439,7 @@ dependencies = [
 
 [[package]]
 name = "rsky-relay"
-version = "0.1.2"
+version = "0.1.3"
 dependencies = [
  "bytes",
  "chrono",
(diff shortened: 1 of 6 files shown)


afbase (commented):
LGTM

PR #261, opened 2026-09-07
fix(syntax): validate the record key segment of an at-uri

afbase:
## Summary

`ensure_valid_at_uri` and `ensure_valid_at_uri_regex` validated the authority and the collection but never validated the trailing segment as a record key, so URIs ending in `.`, `..`, or a malformed percent-escape were accepted.

The length cap was also checked against the fragment-stripped string at the end of the function, which made it both unreachable and blind to an oversized fragment; it now runs against the whole input up front.

Adds the upstream interop syntax corpus as a table-driven suite with exact case counts.

## Test plan

- [ ] `cargo test -p rsky-syntax` passes
- [ ] Every wired interop category passes on both the strict and regex validators
- [ ] `cargo test -p rsky-pds --lib` still passes (only external consumer of this function)


afbase (commented):
LGTM

PR #262, opened 2026-09-07
test(repo): run the upstream mst and commit-proof interop vectors

afbase:
## Summary

`tests/interop_src` was an uninitialized git submodule, so the suite's glob matched nothing and the interop test reported success having run zero cases. This removes the submodule and the harness that depended on it, and vendors the upstream MST and commit-proof vectors directly so the suite has no external checkout requirement.

Three table-driven tests cover common-prefix length, key heights, and the commit proof fixtures. Each asserts an exact case count, so a fixture that fails to load is a failure rather than a silent skip. Fixtures are loaded with `include_str!`, making a missing file a compile error.

## Test plan

- [ ] `cargo test -p rsky-repo` passes
- [ ] Each vector file reports its full expected case count
- [ ] `git submodule status` is empty and no `.gitmodules` remains


afbase (commented):
LGTM

PR #263, opened 2026-09-07
fix(repo): emit the covering proof in commit blocks

afbase:
## Summary

Stacked on #262 — merge that first; this targets it so the diff stays reviewable.

A firehose `#commit` must carry enough MST nodes for a consumer to invert the operation and recover the previous root without holding a copy of the repo. `format_commit` built that set from a path-only walk with no sibling coverage, so the emitted `blocks` could be under-inclusive and the commit not invertible.

Adds `proof_for_key`, `proof_for_left_sib`, `proof_for_right_sib` and `get_covering_proof`, ported from the reference implementation, and uses the covering proof when formatting a commit. Tree mutation is untouched.

## Test plan

- [ ] Every commit-proof vector matches `blocksInProof` by exact set equality
- [ ] All root CIDs in the vectors are unchanged
- [ ] `cargo test -p rsky-repo` passes
- [ ] Reverting `format_commit` to the previous walk fails the proof assertion
- [ ] `cargo check -p rsky-pds -p rsky-wintermute` succeeds

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8497,7 +8497,7 @@ dependencies = [
 
 [[package]]
 name = "rsky-repo"
-version = "0.0.7"
+version = "0.0.8"
 dependencies = [
  "anyhow",
  "async-recursion",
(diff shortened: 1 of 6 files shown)


afbase (commented):
LGTM

PR #264, opened 2026-09-07
test(pds): cover sync 1.1 commit, sync and account event shapes

afbase:
## Summary

There was no end-to-end coverage of `subscribeRepos` or the sync 1.1 event shapes. This adds a suite driving the sequencer, outbox and frame encoder through the real XRPC write endpoints, plus the WebSocket upgrade rejection path through the real route.

Covers `prevData` on every post-genesis commit and its equality with the prior commit's data root, `tooBig` and `rebase` staying false, `ops[].prev` on update and delete, `#sync` and `#account` framing, cursor backfill, and the `applyWrites` operation cap.

One test is `#[ignore]`d with its reason inline: `deactivateAccount` sequences no `#account` event, so a relay is never told the repo went inactive. No source is changed here; that is a follow-up.

## Test plan

- [ ] `cargo test -p rsky-pds --test firehose_tests` passes
- [ ] `cargo test -p rsky-pds --test firehose_tests -- --ignored` reproduces the deactivate gap
- [ ] `cargo build --release -p rsky-pds` succeeds

### DIFF
--- a/rsky-pds/tests/firehose_tests.rs
+++ b/rsky-pds/tests/firehose_tests.rs
@@ -1,0 +1,897 @@
+use futures::StreamExt;
+use lexicon_cid::Cid;
+use rocket::http::{ContentType, Header, Status};
+use rocket::local::asynchronous::Client;
+use rsky_common::time::from_str_to_utc;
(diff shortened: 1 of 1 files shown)


afbase (commented):
LGTM

PR #265, opened 2026-09-07
test(pds): cover the account migration endpoints and round-trip

afbase:
## Summary

Account migration had no integration coverage. This adds a suite over `checkAccountStatus`, `importRepo`, `listMissingBlobs`, `sync.getBlocks`, `sync.getRecord` and activate/deactivate, including a full export-and-import round trip between two in-process instances holding the same DID. CAR fixtures are built in-process.

It also confirms `importRepo` is atomic: a CAR truncated at several points leaves the target untouched.

Eight tests assert protocol-correct behaviour that the server does not yet produce and are `#[ignore]`d with the specific divergence named inline, so they run under `--ignored` and document the gap without failing CI. The largest is `listMissingBlobs` serializing `record_uri` rather than `recordUri`, which a spec-conformant client reads as absent. No source is changed here; those are follow-ups.

## Test plan

- [ ] `cargo test -p rsky-pds --test migration_tests` passes
- [ ] `cargo test -p rsky-pds --test migration_tests -- --ignored` reproduces each named divergence
- [ ] `cargo build --release -p rsky-pds` succeeds

### DIFF
--- a/rsky-pds/tests/migration_tests.rs
+++ b/rsky-pds/tests/migration_tests.rs
@@ -1,0 +1,1013 @@
+//! Account migration: `checkAccountStatus`, `importRepo`, `listMissingBlobs`,
+//! the `sync` proof endpoints an importing server reads, the
+//! activate/deactivate state machine, and the export -> import round trip
+//! between two independent server instances.
+//!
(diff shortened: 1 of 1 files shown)


afbase (commented):
LGTM

PR #267, opened 2026-09-08
ci: skip path-shaped workspace members by glob; drop stale backpressure step

afbase:
Two CI fixes, both ahead of the wintermute hubble-backfill PR (branch `feat/wintermute-hubble-backfill`) so that branch can go green.

## 1. `determine-workspace-members.sh`: skip path-shaped members by glob

`rust.yml` turns each entry of the root `Cargo.toml` `[workspace] members` array into `cargo check|build|test -p <entry>`. That only works when the member path is also the package name. The backfill branch adds a vendored member `vendor/hubble-sync`, which makes cargo fail with:

```
error: package ID specification `vendor/hubble-sync` looks like a file path
```

Changes:

- `SKIP_PACKAGES` entries are now bash glob patterns. Added `vendor/hubble-sync` (the entry that motivated this) and `vendor/*`, so any future vendored crate is skipped without editing the list.
- The skip list is applied in one place, `add_member`, that every member-discovery path goes through: the `members = [...]` array parse, the find-based fallback, and the `rsky*` fallback. The final hardcoded `rsky-*` fallback checks `is_skipped` too. Previously the filter ran once, late, with exact string match only.
- `set -eu pipefail` never enabled pipefail (it set `$1` to `pipefail`); it is now `set -euo pipefail`. The members `grep -o` is guarded with `|| true` so an empty array does not trip it.
- `readarray` (bash 4+) replaced with a `read` loop, and possibly-empty arrays use the `${arr[@]+"${arr[@]}"}` idiom, so the script runs under `set -u` on bash 3.2 as well as GitHub's ubuntu-latest bash 5. The script on `main` exits 127 with `readarray: command not found` under macOS `/bin/bash`.

Verified locally under bash 3.2.57: push path, PR path (`PR_BASE_SHA`/`PR_HEAD_SHA` set to two `main` commits), and both the `.github`-changed "all packages" branch and the changed-files-only branch. With `vendor/hubble-sync`, `vendor/other-crate`, and `cypher/frontend` temporarily added to the members array, all three are logged as `Skipping workspace member:` and none reach `workspace_members=`. The find-based fallback (members not in array form) skips a fake `vendor/fake-crate/Cargo.toml` and `rsky-pdsadmin`; the `rsky*` fallback (no `[workspace]` section) skips `rsky-pdsadmin`. The temporary `Cargo.toml` edits were reverted and are not part of this PR.

## 2. `rust.yml`: remove the "Run rsky-wintermute backpressure test" step

The step ran `cargo test --release -p rsky-wintermute -- --ignored backpressure` with `BACKFILLER_OUTPUT_HIGH_WATER_MARK=100`. The backfill PR deletes the old backfiller and with it `test_backpressure_metric_tracking`, which would leave this step running zero tests.

To be plain about the ordering: that test **still exists on `main`** (`rsky-wintermute/src/backfiller/tests.rs`, `#[ignore]`d), and this step was its only runner in CI. Between this PR merging and the backfill PR merging, that one test is not run in CI. The test itself is not touched here; only the workflow step and its two-line comment are removed. If the gap is unacceptable, merge this PR together with or after the backfill PR.

`rust.yml` still parses (`ruby -ryaml` and `yq`); the remaining `build-and-test` steps are unchanged.

## Notes

- Commits are unsigned (`-c commit.gpgsign=false`; no GPG pinentry in the authoring environment). The author may re-sign before merge.
- No Rust code changes; `.github/` only.

### DIFF
--- a/.github/scripts/determine-workspace-members.sh
+++ b/.github/scripts/determine-workspace-members.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -eu pipefail
+set -euo pipefail
 
 # Get GitHub event variables from environment
 EVENT_NAME="${GITHUB_EVENT_NAME}"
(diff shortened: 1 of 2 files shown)


afbase (commented):
LGTM

PR #269, opened 2026-09-08
fix(relay): make rsky-relay clippy-clean under Rust 1.98

afbase:
## Summary

Makes `rsky-relay` clippy-clean under Rust 1.98 while staying clean under the pinned 1.86 toolchain. Lint-only: no behavior changes, no test semantics changed, `rust-toolchain` untouched.

This prepares for the toolchain bump to 1.98 in `feat/wintermute-hubble-backfill`, where `cargo clippy -p rsky-relay --all-targets` currently fails with 62 errors because the crate uses `#![deny(warnings, clippy::all, clippy::pedantic, clippy::nursery, ...)]`.

## Lint categories and how each was resolved

| Lint | Count | Resolution |
|---|---|---|
| `result_large_err` (fn + closure) | 33 | Real fix: boxed the large payloads. `tungstenite::Error` (136 bytes, over the 128-byte limit) is now `Box<tungstenite::Error>` in `crawler::ConnectionError`, `publisher::ConnectionError`, and the `HandshakeResult` alias. The publisher `Handshake` variant carried the mid-handshake stream (~1440 bytes) and is now `Box<ServerHandshakeError>`. Manual `From` impls keep every `?` site working; `crawler/client.rs` (vendored tungstenite `connect`) gained explicit `map_err(Error::from)` on the few `io::Error`/`ToStrError`/`InvalidUri` conversions. |
| `large_enum_variant` | 4 | Fixed by the same boxing (`ConnectionError`, `WorkerError`, `ManagerError`, `RelayError` all shrink). |
| `collapsible_if` | 20 | Every suggestion required let-chains (Rust 1.88+), which the pinned 1.86 toolchain cannot compile. Declared `rust-version = "1.86"` in `rsky-relay/Cargo.toml`; clippy reads that as MSRV and no longer suggests them. |
| Duration smaller-unit (`Duration::from_mins`/`from_hours`) | 3 | Same MSRV declaration (those constructors are 1.91+). |
| `string_to_string` removed (covered by `implicit_clone`) | 2 | Removed from the `deny` list and the `cfg_attr(test, allow(...))` list. |
| `match` used for equality check | 1 | Rewrote the query-param `match` in `Server::host_status` as `if key.as_ref() == "hostname"`. |
| Unfulfilled `#[expect(clippy::redundant_clone)]` | 1 | Removed from the `expect` on `connect_with_config`. Clippy 1.86 still reports that clone as redundant (false positive fixed upstream; `uri` is reassigned per redirect while `parts` stays borrowed), so the single `let` carries a targeted `#[allow(clippy::redundant_clone)]` with a comment. |
| `missing_const_for_fn` | 1 | Silenced by the MSRV declaration (the suggestion depended on newer const stabilizations). |

## Verification

- `cargo clippy -p rsky-relay --all-targets` (pinned 1.86): 0 errors, 0 warnings from `rsky-relay`
- `RUSTUP_TOOLCHAIN=1.98 cargo clippy -p rsky-relay --all-targets`: 0 errors, 0 warnings from `rsky-relay`
- `cargo test -p rsky-relay` (1.86): 120 passed, 0 failed
- `cargo fmt --all -- --check`: clean

Remaining warnings in the clippy output come from `rsky-common`, `rsky-identity`, and `rsky-crypto`, plus a pre-existing cargo manifest warning about `src/main.rs` being shared by the `rsky-relay` and `rsky-relay-labeler` bin targets; none are introduced here.

Not addressed (pre-existing, out of scope): with `--features labeler`, rustc 1.98's stricter `dead_code` analysis flags five `cfg(not(feature = "labeler"))`-only types in `server/types.rs` and `validator/event.rs`. That build is clean on 1.86 and CI does not exercise the feature.

## Notes

Commits are unsigned (GPG pinentry unavailable in the authoring environment).

### DIFF
--- a/rsky-relay/Cargo.toml
+++ b/rsky-relay/Cargo.toml
@@ -3,6 +3,10 @@ name = "rsky-relay"
 version = "0.1.3"
 default-run = "rsky-relay"
 edition = "2024"
+# Keep in sync with the toolchain pin in `rust-toolchain`. Clippy reads this as
+# its MSRV so it does not suggest syntax/APIs (let chains, `Duration::from_mins`)
+# that the pinned toolchain cannot compile.
+rust-version = "1.86"
 
 [dependencies]
 # external
--- a/rsky-relay/src/crawler/client.rs
+++ b/rsky-relay/src/crawler/client.rs
@@ -8,7 +8,7 @@ use socket2::{Domain, Protocol, Socket, Type};
 use tungstenite::Connector;
 use tungstenite::client::{IntoClientRequest, uri_mode};
 use tungstenite::client_tls_with_config;
-use tungstenite::error::{Error, Result, UrlError};
+use tungstenite::error::{Error, UrlError};
 use tungstenite::handshake::client::Request;
 use tungstenite::protocol::WebSocketConfig;
 use tungstenite::stream::{Mode, NoDelay};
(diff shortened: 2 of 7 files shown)


afbase (commented):
LGTM

PR #271, opened 2026-09-09
fix(lexicon): decode subscribeRepos#commit blobs as cid-links

afbase:
## Production symptom

On production wintermute, every `#commit` frame from a `tranquil` PDS (e.g. `airlock.ltd`) that carries a blob fails to decode with

```
Mismatch { expect_major: 3, byte: 216 }
```

and the whole event is dropped — hundreds per hour. `expect_major: 3` is CBOR text-string; byte `216` is `0xD8`, the start of a CBOR tag.

## Cause

`rsky-lexicon`'s `SubscribeReposCommit` declared `blobs: Vec<String>`. The lexicon `com.atproto.sync.subscribeRepos#commit` defines `blobs` as:

```json
"blobs": { "type": "array", "items": { "type": "cid-link" } }
```

In DAG-CBOR a `cid-link` is tag 42 (`0xD8 0x2A`) wrapping the binary CID, not a text string. The reference PDS always sends an empty `blobs` array, so the mismatch was invisible until a PDS that populates the field appeared on the firehose.

## Change

- `blobs` is now `Vec<Cid>` (the same `Cid` type used for `commit`, `prev`, `prevData`), with `#[serde(default)]` so a missing field decodes as empty.
- New `deserialize_cid_vec_v1` decodes each element leniently: tag-42 CID links (what the lexicon requires and what the reference PDS / tranquil emit), **and** text-string CIDs, because rsky-pds itself has been emitting `blobs` as strings (`blob.to_string()` in `subscribe_repos.rs`) and those frames exist in the wild — relays and consumers must keep parsing them. Raw untagged CID bytes are accepted too, mirroring `deserialize_cid_v1`. It drives the deserializer through `deserialize_any` exactly like the existing `Tagged<ByteBuf>` helpers, so a tag-42 link arrives via `visit_newtype_struct` under both `serde_ipld_dagcbor` and `serde_cbor`.
- Serialization now emits proper tag-42 CID links. `rsky-pds`'s firehose passes its `Vec<Cid>` straight through instead of stringifying, so rsky-pds now emits a conformant `blobs` field.
- `rsky-lexicon` bumped `0.10.3 -> 0.11.0` (public field type change); root `Cargo.toml` constraint and `Cargo.lock` updated.

## Tests (rsky-lexicon)

- `commit_decodes_blobs_encoded_as_cid_links` — hand-built DAG-CBOR body with two tag-42 links. **Fails on `main`** with the exact production error `Mismatch { expect_major: 3, byte: 216 }`; passes here.
- `commit_decodes_blobs_encoded_as_text_strings` — the rsky-pds legacy encoding still decodes.
- `commit_decodes_missing_blobs_as_empty` — `#[serde(default)]` (also failed on `main`: `missing field 'blobs'`).
- `commit_round_trips_blobs_as_cid_links` — serializes two blobs, asserts the `D8 2A 58 <len> 00 <cid>` byte sequence is present for each and that the value re-reads as `Ipld::Link`, then round-trips equal.

## Verification

- `cargo test -p rsky-lexicon` — 57 passed
- `cargo check --all-targets -p rsky-pds -p rsky-relay -p rsky-firehose -p rsky-wintermute -p rsky-jetstream-subscriber -p rsky-labeler` — clean
- `cargo clippy -p rsky-lexicon --all-targets` — only the 4 pre-existing `large_enum_variant` warnings, none in changed code
- `cargo fmt --all -- --check` — clean

Note: `rsky-relay` has its own `validator::event::SubscribeReposCommit` with `blobs: Vec<String>` (marked deprecated); it is not touched here and likely has the same latent decode failure — worth a follow-up.

Commits are unsigned (GPG pinentry unavailable in this environment).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8320,7 +8320,7 @@ dependencies = [
 
 [[package]]
 name = "rsky-lexicon"
-version = "0.10.3"
+version = "0.11.0"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ rand_core = "0.6.4"
 secp256k1 = { version = "0.28.2", features = ["global-context", "serde", "rand", "hashes","rand-std"] }
 serde_json = { version = "1.0.96",features = ["preserve_order"] }
 rusqlite = { version = "0.36", features = ["bundled"] }
-rsky-lexicon = {path = "rsky-lexicon", version = "0.10.0"}
+rsky-lexicon = {path = "rsky-lexicon", version = "0.11.0"}
 rsky-identity = {path = "rsky-identity", version = "0.2.0"}
 rsky-crypto = {path = "rsky-crypto", version = "0.2.1"}
 rsky-syntax = {path = "rsky-syntax", version = "0.1.1"}
--- a/rsky-lexicon/Cargo.toml
+++ b/rsky-lexicon/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rsky-lexicon"
-version = "0.10.3"
+version = "0.11.0"
 edition = "2021"
 publish = true
 description = "Bluesky API library"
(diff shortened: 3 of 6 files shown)


afbase (commented):
LGTM

PR #272, opened 2026-09-09
fix(wintermute): drop stale live replays of commits before enqueue

afbase:
## Incident

Production wintermute subscribes to three relays at once (`RELAY_HOSTS=bsky.network,atproto.africa,blacksky.app`) with no cross-source dedupe, and atproto.africa runs ~37 h behind, so every commit carried by more than one relay was applied again a day or two late. The post upsert is rev-gated so existing rows survived, but a post the user had deleted in between had no row to gate on and the late create re-inserted it with its original rev (observed: deleted Sept 6, back on Sept 8 with a Sept 5 rev), and every `record` row got its `indexedAt` bumped by the replay.

## Design

The sync 1.1 tracker already judged these frames `stale` but only skipped advancing its own state; the jobs were still enqueued and indexed. This PR makes the tracker's existing two-generation cache the gate:

- **Where the gate sits.** `IngesterManager::enqueue_live_commit` (extracted from the live loop in `ingester/mod.rs`) consults `Sync11Handle::is_stale_replay(did, rev)` *before* the CAR is parsed and before anything is enqueued. When a rev is known for the DID and the commit's rev is not newer (`!rev_newer`), every op of the commit (creates, updates, deletes) is dropped, nothing goes to fjall, the tracker is not sent a message, `ingester_live_commits_dropped_total{reason="stale_replay"}` is incremented, and a `debug` line carries did/rev/seq/host. Non-stale commits follow the unchanged flow: enqueue to fjall, then tell the tracker.
- **Why before enqueue.** The indexer's per-row rev gate can only protect rows that exist; a deleted record has no row, so a late create simply re-inserts it. The commit has to be refused as a whole, upstream of the queue.
- **The shared view.** Rather than a second mirror map, the tracker's `Cache` itself now lives behind `Arc<RwLock<Cache>>` (`sync11::KnownRevs`), shared with every `Sync11Handle`. The ingester does one read-locked, non-promoting hash lookup (`Cache::peek_rev`), no allocation, no Postgres, no await on the tracker task. Because it *is* the cache, it is bounded to exactly the cache's budget (2 x 150k entries) and is updated by every path that changes what the tracker holds: a recorded commit, the batch `repo_sync` warm-up load on cache miss, and a forget on `#account` write-off or `#sync`. The tracker pays one uncontended lock acquire per cache op; no extra memory.
- **Cold-cache caveat** (documented on `KnownRevs` and `enqueue_live_commit`, and in the README). A repo the tracker has not touched since the process started (or that rotated out of the cache) has no entry, so its next commit passes even if it is a replay; the tracker then loads the persisted row and judges it `stale`, and the entry is present from then on. The gate therefore never drops anything newer than what we hold, but it cannot catch the first replay per repo after a restart, nor a replay that lands inside the tracker's channel lag behind the original.
- `#identity`, `#account` and `#sync` frames are not gated.

## Metric

`ingester_live_commits_dropped_total{reason="stale_replay"}` (added in `metrics.rs`; documented in the README metrics list and the sync 1.1 section).

## Tests

- `sync11`: pure gate (`is_stale_replay`): newer passes, equal dropped, older dropped, unknown DID passes, malformed revs by equality; `Cache::peek_rev` reads both generations without promoting; the handle's view updates when the tracker records a commit (and not on a stale one), when the batch warm-up loads rows from the in-memory `MemStore`, and clears on write-off; a `spawn`ed handle shares the actor's view.
- `ingester`: `a_stale_replay_enqueues_nothing_and_is_counted` drives a real fjall `Storage` and a hand-driven `Tracker` sharing a handle: the first commit enqueues 2 jobs and the tracker is told; the same commit again (and an older one) enqueues nothing, sends nothing to the tracker, and bumps the counter by exactly 2; a newer commit enqueues again.
- `MemStore` moved to `sync11::testing` (`#[cfg(test)]`) so the ingester test can use it.

`cargo fmt --all -- --check`: clean. `cargo clippy -p rsky-wintermute --all-targets`: zero diagnostics for the crate (this also clears two pre-existing `manual_checked_ops` warnings in the `plc_import` and `label_sync` bins). `cargo test -p rsky-wintermute`: 220 passed, 2 ignored, 31 failed; the 31 are the pre-existing Postgres-backed indexer/ingester tests that need a migrated DB and are unrelated to this change.

Commits are unsigned (GPG pinentry unavailable in this environment).

### DIFF
--- a/rsky-wintermute/README.md
+++ b/rsky-wintermute/README.md
@@ -216,6 +216,7 @@ Prometheus metrics are exposed at `http://localhost:9090/metrics`:
 - `ingester_sync11_commits_total{outcome}` - Live `#commit` frames checked against stored sync state: `applied`, `first_seen`, `stale`, `lax`, `desync`, `no_data`
 - `ingester_sync11_sync_events_total{outcome}` - Live `#sync` frames: `resync` or `unchanged`
 - `ingester_sync11_resyncs_requested_total{reason}` - Repos handed to backfill for a full resync: `prev_mismatch` or `sync_event`
+- `ingester_live_commits_dropped_total{reason}` - Live `#commit` frames dropped whole before enqueue: `stale_replay` (the rev does not advance what the sync 1.1 tracker holds for the repo)
 - `ingester_identity_tasks_in_flight` - Identity/account/sync tasks running (bounded by `IDENTITY_EVENT_CONCURRENCY`)
 - `ingester_identity_task_timeouts_total{kind}` / `ingester_identity_tasks_shed_total{kind}` - Tasks abandoned at the deadline; events dropped because every permit was busy
 - `wintermute_rss_bytes` / `wintermute_rss_peak_bytes` - Resident set size and its high-water mark (Linux; 0 elsewhere)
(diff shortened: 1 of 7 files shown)


afbase (commented):
LGTM

PR #284, opened 2026-09-17
fix(pds): preserve NUX state and extra preference fields

rishibalakrishnan:
Preserve NUX completion state and unrecognized fields in app-state preferences, progress guides, and muted words so preference saves do not discard client data. Add flattened JSON fields alongside typed NUX support and SQLite put/get regression coverage; retain the lexicon patch bump and remove the PDS version bump.

Validation: formatting, all 59 lexicon tests, all 9 preference-store tests, and the lexicon release build pass. The storage regression checks exact stored JSON and returned preferences for absent, empty, and populated NUX arrays, beta-user state, progress fields, muted-word expiry, and future fields.

Strict Clippy remains blocked by four existing `large_enum_variant` diagnostics in unchanged definitions. Lexicon coverage is 87.76% lines / 77.78% functions, below the required 100%. Current main is already included; the CI fixes mentioned in review have not yet landed there.

- [ ] Resolve or review existing lint and coverage gate failures.
- [ ] Refresh against main after the pending CI fixes land and verify PR CI.
- [ ] Verify an HTTP preferences write/read through the running service.


rudyfraser (changes requested):
Thanks Rishi, the bug is real and the change is right as far as it goes. I don't want to merge it in this shape though, because it patches one symptom of a storage problem that keeps eating other fields.

What works:

- `putPreferences` stores `serde_json::to_string(&RefPreferences)` (`rsky-pds/src/actor_store/preference/mod.rs:69`), so any lexicon field the typed struct lacks is dropped on write. `nuxs` was one of them, and adding it stops completed announcements from coming back.
- `Nux` matches the lexicon (`id` and `completed` required, `data` and `expiresAt` optional, camelCase).
- The round-trip test covers absent, empty and populated arrays.
- The rsky-lexicon and workspace dependency bumps are right.

What I'd like changed:

- The reference PDS stores each preference untouched (`valueJson: JSON.stringify(value)` in `packages/pds/src/actor-store/preference/transactor.ts`). rsky re-encodes through typed structs with no catch-all, so every field added to a known `$type` after the struct was written is lost. Today that includes `isBetaUser` on this same struct, `id`, `actorTarget` and `expiresAt` on `MutedWord` (muted-word expiry is lost on every save), and the unspecced progress fields the lexicon says `bskyAppProgressGuide` carries.
- The fix that closes the class is the same size as this PR: give the preference structs a catch-all, e.g.

  ```rust
  #[serde(flatten)]
  pub extra: serde_json::Map<String, serde_json::Value>,
  ```

  on `BskyAppStatePref`, `MutedWord` and `BskyAppProgressGuide`. Keeping the typed `nuxs` field on top is fine if the PDS wants to read it, but it should not be what preserves it.
- Please add the regression test on the path that lost the data: a put and get through `rsky-pds/src/actor_store/preference/mod.rs` (there is a test module there) with a `bskyAppStatePref` carrying `nuxs`.
- Drop the rsky-pds bump. No rsky-pds code changes here and its `rsky-lexicon` dependency comes from the workspace table. It will also conflict with the 1.14.2 landing on main shortly.

Not blocking: `id` (100), `data` (3000 chars / 300 graphemes) and `nuxs` (100 items) have length limits in the lexicon that rsky does not check. We don't check them for other preferences either, so that's a follow-up issue.

On CI: both failures are main's, not yours. `sync_endpoints_reject_a_request_missing_a_required_parameter` fails on main too and the `coverage-pds` runner dies from disk exhaustion on every recent run. Fixes for both are about to land on main; a rebase after that should go green without changes on your side.

rudyfraser (approved):
This does what I asked, thanks Rishi. Approving.

Checked locally on c70b2342:

- rsky-lexicon: 59 tests pass. The round-trip tests hold `$type` outside the catch-all as expected for the internally tagged enum, so nothing is emitted twice.
- rsky-pds preference store: 9 tests pass, including the new put/get that compares the stored JSON byte for byte.
- No struct-literal construction of the three structs anywhere in the workspace, so the added `extra` fields break nothing.
- The four `large_enum_variant` diagnostics are in feed, chat and sync definitions this PR does not touch.

One follow-up, not blocking here: the same drop still applies to every other struct reachable from `RefPreferences` (`SavedFeed`, `SavedFeedsPrefV2`, `FeedViewPref`, `ThreadViewPref`, `ContentLabelPref`, `LabelersPrefItem`, `Nux`, and the rest). A second PR giving those the same `#[serde(flatten)] extra` closes the class for good; a small helper test looping over one sample per `$type` would keep it honest.

On merging: the CI fixes are committed on main locally and go out with the next push. Once they land, rebase and CI should be green; I'll merge after that.

PR #285, opened 2026-09-19
fix(rsky-video): handle duplicate upload jobs in playback lookup

rishibalakrishnan:
Repeated uploads of the same video can make the playlist and thumbnail lookup return multiple rows, causing HTTP 500 responses. Aggregate matching jobs into one result while preserving private status if any job is private, including after a later public upload; add regression tests and bump rsky-video to 0.2.1.

Test migrations run once per process so concurrent tests work on a fresh database. Each test retains its own connection pool.

Validation:
- Formatting and diff checks pass.
- All three database regression tests pass concurrently against a fresh local PostgreSQL instance: `TEST_DATABASE_URL=... cargo test -p rsky-video db::tests -- --ignored --test-threads=3`.
- Standard crate suite: 12 passed, 10 ignored; the three database tests were separately exercised above.
- Clippy with warnings denied remains blocked by four existing warnings in unchanged PDS/signing code.
- Coverage and release build remain unverified.

- [ ] Resolve existing lint blockers.
- [ ] Run the coverage gate and release build.


rudyfraser (changes requested):
The fix is right: `video_mappings` is `UNIQUE (did, cid)` so the mapping side is one row, and `bool_or` over the joined jobs makes the lookup single-row again while failing closed on privacy. The index on `(did, video_cid)` is what that join wanted anyway.

I ran the three Postgres tests on a build box against a throwaway Postgres 17. Two results:

- On a warm database (migrations already applied) all three pass, and clippy is clean for the touched file (the four remaining warnings are in `pds/` and `signing/`, untouched here).
- On a fresh database, run as documented (`cargo test -p rsky-video -- --ignored` with `TEST_DATABASE_URL`), two of the three fail:

  ```
  duplicate key value violates unique constraint "pg_type_typname_nsp_index"
  duplicate key value violates unique constraint "pg_class_relname_nsp_index"
  ```

  Each test calls `run_migrations` and they run concurrently, so the `CREATE TABLE IF NOT EXISTS` statements race inside Postgres. With `--test-threads=1` all three pass. Please run the migrations once per process (a `tokio::sync::OnceCell` in the test module is enough) so the tests pass the first time someone follows the instructions.

One thought, not blocking: the mapping is written by the Bunny webhook with the finished job's own `video_guid`, so `LEFT JOIN videos.video_jobs j ON j.bunny_video_id = m.bunny_video_id` would pick exactly the job that produced the mapping, and a video re-uploaded publicly after a private attempt would be served as its latest upload says. `bool_or` keeps it private forever in that case. Either is defensible; the fail-closed choice is fine if that's the intent, just say so in the comment.

The unrelated `transcode::` and `bunny::` ignored tests fail on that box for lack of ffmpeg and live credentials; not this PR's concern.

rudyfraser (approved):
Approving. Verified on 50c1aeb4 against fresh Postgres 17 instances on a build box: the three database tests pass concurrently on three separate fresh databases and again on a warm one; the crate suite passes (12, 10 ignored); clippy has nothing in the touched file. The `OnceCell` is the right shape, and the Cargo.lock bump is in now too.

The comment now states the fail-closed intent for a later public re-upload, which answers my question. Thanks Rishi.

